PR #1, opened 2025-02-07
LLK test infra PR

ldjurovicTT:
# Initial pull request for architecture and stack independent test infra for LLKs

## Overview
The idea is to set up infrastructure that will allow users to test LLK code outside of TT run stack such as tt-metal. For now it runs on WH and BH machines

It consists of few parts that are all located in `tests/` folder
- firmware -> taken mostly from tt-metal and contains files needed to support infra that are not LLK code ( l1_address_map, tensix_types etc... ). They are divided into common and specialized files per architecture
- helpers -> code supporting infra that doesn't come from tt-metal. It contains ckernel helper functions device memory map and a header called params.h that is used in parametrizing tests. Besides that it contains .ld files for all TRISC cores and BRIS, and trisc.cc and brisc.cc that hold main functions for those cores ( mostly taken from tt-metal)
- sources -> c++ test files that contain pure LLK code with no API
- python_tests -> contain pytest source code for each cpp test. Inside of them there is also helpers folder that contains all of python supporting code needed for infra to run like stimuli_generator, functions that talk to device, test configurations ... and run_test.sh script
- In `tests/` root there is also setup_env.sh which is used to get `tt-lens` and `sfpi` when setting up test infra.

## Usage

```
git clone https://github.com/tenstorrent/tt-llk
cd tt-llk
git checkout ldjurovic/llk_tests
cd tests
source setup_env.sh
```
After that You can navigate to python_tests and from there You have 2 options for running tests.
First one is to just use `pytest <testname>`, and the second one is by using `run_test.sh` in one of ways it can be used:
```
./run_test.sh --repeat <number_of_repeats> --test <test_name> [--log <log_file>]
./run_test.sh --all 
```
It would be very beneficial if reviewers could provide feedback on both architecture and implementation of this test infra.
All other informations can be found on infra's docs page located [here](https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/563315664/LLK+test+envirionment)


tests/Makefile
```diff
@@ -0,0 +1,133 @@
...
+RMDIR := rm -rf
+
+# Tool paths
+GXX=$(TOOL_PATH)/riscv32-unknown-elf-g++
+OBJDUMP=$(TOOL_PATH)/riscv32-unknown-elf-objdump
+OBJCOPY=$(TOOL_PATH)/riscv32-unknown-elf-objcopy
+READELF=$(TOOL_PATH)/riscv32-unknown-elf-readelf
+
+# GCC options
+OPTIONS_ALL=-O3 -mabi=ilp32 -std=c++20 -g -ffast-math
```
fvranicTT, line 13:
```suggestion
OPTIONS_ALL=-O3 -mabi=ilp32 -std=c++20 -g -ffast-math -fprofile-arcs -ftest-coverage
```
To enable coverage. Let's see if that's enough for now.

tests/sources/eltwise_binary_test.cc
fvranicTT, line 1: Should this dir's name change from sources to something else? It sounds too vague to me.

tests/sources/eltwise_binary_test.cc
```diff
@@ -0,0 +1,88 @@
+#include <cstdint>
+#include <cstdio>
+
+#include "llk_defs.h"
+#include "ckernel.h"
+
+// Globals
+uint32_t unp_cfg_context = 0;
+uint32_t pack_sync_tile_dst_ptr = 0;
+volatile uint32_t tt_l1_ptr l1_buffer[16] __attribute__ ((section (".text#"))) __attribute__ ((aligned (16)));
```
fvranicTT, line 10:
```suggestion
volatile std::uint32_t tt_l1_ptr l1_buffer[16] __attribute__ ((section (".text#"), aligned (16)));
```
But more importantly, why do we want this array in the .text section? It sounds a bit odd to be honest. Could you elaborate on that? Even better, including an explanation in form of a comment would be nice.

tests/sources/eltwise_unary_datacopy_test.cc
```diff
@@ -0,0 +1,107 @@
...
+#include "llk_pack_common.h"
+#include "params.h"
+
+volatile uint32_t* buffer_Dest = (volatile uint32_t*)0x1c000;
+void run_kernel()
+{
+    for(int i = 0; i < 16*16*4; i++)
+    {
+        buffer_Dest[i] = 0xdeadbeef;
+    }
```
fvranicTT, line 87:
```suggestion
    std::fill(buffer_Dest, buffer_Dest + 16 * 16 * 4, 0xdeadbeef);
```

tests/sources/eltwise_unary_sfpu_test.cc
```diff
@@ -0,0 +1,121 @@
...
+// #define PACK_DEST_FORMAT DATA_FORMAT
+
+
+volatile uint32_t* buffer_Dest = (volatile uint32_t*)0x1c000;
+void run_kernel()
+{
+    for(int i = 0; i < 16*16*4; i++)
+    {
+        buffer_Dest[i] = 0xdeadbeef;
+    }
```
fvranicTT, line 101:
```suggestion
    std::fill(buffer_Dest, buffer_Dest + 16 * 16 * 4, 0xdeadbeef);
```

tests/python_tests/run_test.sh
```diff
@@ -0,0 +1,84 @@
...
+        echo "Running test: $test_name (Iteration $i)" | tee -a "$log_file"
+        pytest --color=yes -v "$test_name" | tee -a "$log_file" 2>&1
+    else
+        echo "Running test: $test_name (Iteration $i)"
+        pytest --tb=short --color=yes -v "$test_name"
+    fi
+
+    result=$?
+    
+    if [ $result -eq 0 ]; then
```
fvranicTT, line 79:
```suggestion
    if [[ $result -eq 0 ]]; then
```

tests/python_tests/test_eltwise_binary.py
```diff
@@ -0,0 +1,74 @@
...
+    test_config = {
+        "input_format": format,
+        "output_format": format,
+        "testname": testname,
+        "dest_acc": dest_acc,
+        "mathop": mathop
+    }
+
+    make_cmd = generate_make_command(test_config)
+    run_shell_command(f"cd .. && {make_cmd} >/dev/null")
```
fvranicTT, line 45:
```suggestion
    run_shell_command(f"cd .. && {make_cmd}")
```
No need for `> /dev/null` since you have `capture_output` set to `False`. Applies for other invocations where you set redirection to /dev/null.

tests/firmware/riscv/wormhole/l1_address_map.h
```diff
@@ -0,0 +1,161 @@
...
+#pragma once
+
+#include <cstdint>
+
+// Aux variable used to align addresses to platform specific width. WH requires 32B alignment.
+#define NOC_ADDRESS_ALIGNMENT (32)
+
+namespace l1_mem {
+
+struct mailbox_type {
```
fvranicTT, line 13: should be an enum class.

tests/Makefile
ldjurovicTT: Have You found out which compile flags lead to overflow in sections and which are ok to be added ?

tests/sources/eltwise_unary_datacopy_test.cc
ldjurovicTT: Implemented

tests/python_tests/helpers/utils.py
```diff
@@ -0,0 +1,141 @@
...
+def calculate_read_words_cnt(format, sfpu=False):
+
+    if format not in format_sizes:
+        raise ValueError(f"Unsupported format: {format}")
+
+    if sfpu: # for now just for 16 bit formats
+        if(format in ["Float32", "Int32"]):
+            return 256
+        else:
+            return 128
```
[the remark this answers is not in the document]
ldjurovicTT: Implemented

tests/Makefile
```diff
@@ -0,0 +1,124 @@
...
+	$(GXX) $(OPTIONS_ALL) $(OPTIONS_LINK) $^ -T$(LINKER_SCRIPTS)/memory.$(CHIP_ARCH).ld -T$(LINKER_SCRIPTS)/trisc1.ld -T$(LINKER_SCRIPTS)/sections.ld -o $@
+$(OUTPUT_ELFS)/$(testname)_trisc2.elf: $(BUILD_DIR)/tmu-crt0.o $(BUILD_DIR)/main_pack.o $(BUILD_DIR)/$(testname)_pack.o $(LINK_OBJECTS)
+	$(GXX) $(OPTIONS_ALL) $(OPTIONS_LINK) $^ -T$(LINKER_SCRIPTS)/memory.$(CHIP_ARCH).ld -T$(LINKER_SCRIPTS)/trisc2.ld -T$(LINKER_SCRIPTS)/sections.ld -o $@
+
+# Building brisc.elf for BRISC core
+$(OUTPUT_ELFS)/brisc.elf: $(BUILD_DIR)/tmu-crt0.o $(BUILD_DIR)/brisc.o
+	$(GXX) $(OPTIONS_ALL) $(OPTIONS_LINK) $^ -T$(LINKER_SCRIPTS)/memory.$(CHIP_ARCH).ld -T$(LINKER_SCRIPTS)/brisc.ld -T$(LINKER_SCRIPTS)/sections.ld -o $@
+
+#compiling _test.cc to .o
+$(BUILD_DIR)/$(testname)_unpack.o: sources/$(testname).cc
```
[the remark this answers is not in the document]
ldjurovicTT: For now I think Iw should stay as is since it is idea to compile only one test to ELF at once, not all of them

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,72 @@
...
+        elif stimuli_format == "Float32":
+            write_to_device("0,0", buffer_A_address, pack_fp32(buffer_A_tile))
+            write_to_device("0,0", buffer_B_address, pack_fp32(buffer_B_tile))
+        
+        buffer_A_address += 4096
+        buffer_B_address += 4096
+        
+        
+
+def get_result_from_device(format,read_data_bytes,sfpu=False):
```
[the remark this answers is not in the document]
ldjurovicTT: Veeery nice and clean

tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -0,0 +1,79 @@
+import pytest
+import torch
+import os
+from helpers import *
+
+def generate_golden(operand1,format):
+    return operand1
+
+# @pytest.mark.parametrize("format",  ["Float32","Bfp8_b", "Float16_b", "Float16", "Int32"])
```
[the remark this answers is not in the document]
ldjurovicTT: Yes, paramterizing test name turned out to be very bad idea

tests/python_tests/helpers/unpack.py
```diff
@@ -0,0 +1,90 @@
...
+
+        bfloat16_values.append(((-1)**sign)*(int_value+fract_value))
+
+    return bfloat16_values
+
+def unpack_bfp8_b(bfp8_block,sfpu=False):
+
+    if not sfpu:
+        exponents = bfp8_block[:64]
+        reversed_exponents = revese_endian_chunk(exponents)
```
fvranicTT, line 74: This line can be extracted from the if/else, making it smaller and easier to digest. So we extract `reversed_exponents` outside of if/else. Also, we could reverse mantissas right away, right? And just extract parts that we need:
```python
def unpack_bfp8_b(bfp8_block, sfpu=False):

    if not sfpu:
        exponents = bfp8_block[:64]
        mantissas = bfp8_block[64:]
    else:
        exponents = bfp8_block[:16]
        mantissas = bfp8_block[16:272]

    reversed_exponents = revese_endian_chunk(exponents)  # Reverse exponents once
    reversed_mantissas = revese_endian_chunk(mantissas)  # Reverse mantissas once

    num_blocks = len(reversed_exponents)
    values_per_block = 16
    
    # Pre-allocate the list for bfloat16 values
    bfloat16_values = [None] * num_blocks * values_per_block

    index = 0
    for i in range(num_blocks):
        exponent = reversed_exponents[i]
        bfp8_mantissas = reversed_mantissas[i * 16:(i + 1) * 16]
        block_bfloat16_values = bfp8_to_float_block(exponent, bfp8_mantissas)

        for value in block_bfloat16_values:
            bfloat16_values[index] = value
            index += 1

    return torch.tensor(bfloat16_values, dtype=torch.bfloat16)
```

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,85 @@
...
+        end_index = start_index + TILE_SIZE
+
+        if end_index > len(buffer_A) or end_index > len(buffer_B):
+            raise IndexError("Buffer access out of bounds")
+
+        buffer_A_tile = buffer_A[start_index:end_index]
+        buffer_B_tile = buffer_B[start_index:end_index]
+
+        if stimuli_format == "Float16_b":
+            write_to_device("0,0", buffer_A_address, pack_bfp16(buffer_A_tile))
```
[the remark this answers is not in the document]
ldjurovicTT: Great! Implemented

tests/python_tests/run_test.sh
```diff
@@ -0,0 +1,84 @@
...
+        ((pass_count++))
```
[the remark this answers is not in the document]
ldjurovicTT: For now it counts wrong. It should increment passed or failed test count for every combination inside of pytest. If You run `./run_test.sh --repeat 3 --test test_eltwise_unary_datacopy.py  ` It will print 
Summary:
Passed: 
Failed: 3

since some of instances fail. Rather we would want something like

Passed: 20
Failed: 5

tests/python_tests/test_eltwise_binary.py
ldjurovicTT: It reduces amount of stuff printed in terminal. For example messages like this:

build/tmu-crt0.o
build/main_unpack.o
build/eltwise_unary_datacopy_test_unpack.o
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libstdc++.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libm.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
build/tmu-crt0.o
build/main_math.o
build/eltwise_unary_datacopy_test_math.o
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libstdc++.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libm.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
build/tmu-crt0.o
build/main_pack.o
build/eltwise_unary_datacopy_test_pack.o
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libstdc++.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libm.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
build/tmu-crt0.o
build/brisc.o
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libstdc++.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libm.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libc.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/../../../../riscv32-unknown-elf/lib/libgloss.a
/localdev/ldjurovic/tt-llk/tests/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.2.0/libgcc.a

won't be printed to terminal

fvranicTT: Ah, you are right. I was missing `stdout=subprocess.DEVNULL` parameter in the subprocess.run. Please add it, and then you won't need to use `/dev/null` explicitly.

fvranicTT: We do however want to keep the stderr prints.

tests/firmware/riscv/blackhole/tensix_types.h
```diff
@@ -0,0 +1,332 @@
...
+#endif
+
+// Voluntary FIFO alignment so that we can pack fifo address down to 16 bits in the command.
+// At 8, we can cover 16MB. The upper limit is 15 because the command queue is at 32K (by default)
+// Even though the command queue address never goes into a command, assertions complain if it is misaligned.
+// (Hardware only requires 16B alignment.)
+static constexpr std::uint32_t FIFO_BASE_ADDRESS_ALIGN_BITS = 9;
+static constexpr std::uint32_t FIFO_BASE_ADDRESS_ALIGN = 1 << FIFO_BASE_ADDRESS_ALIGN_BITS;
+
+enum class DataFormat : std::uint8_t {
```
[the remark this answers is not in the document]
fvranicTT: i.e.
```cpp
enum class DataFormat : std::uint8_t {
    Float32 = 0,
    Float16 = 1,
    Bfp8 = 2,
    Bfp4 = 3,
    Tf32 = 4,
    Float16_b = 5,
    Bfp8_b = 6,
    Bfp4_b = 7,
    Int32 = 8,
    UInt16 = 9,
    Lf8 = 10,
    Bfp2 = 11,
    Int8 = 14,
    Bfp2_b = 15,
    UInt32 = 24,
    Fp8_e4m3 = 26,    // Not a valid HW encoding, it is Lf8 encoding + extra 5th bit set to specify Lf8 with E4M3
    UInt8 = 30,
    Uint8 = 129,      // Not a valid HW enum value, but useful to have it here for SW
    testMan7 = 0x82,  // intermediate format for testing: 7bit mantissa (6+hidden)
    testMan2 = 0x8A,  // intermediate format for testing: 2bit mantissa (2+hidden)
    Invalid = 0xff
};
```

tests/firmware/riscv/common/risc_attribs.h
```diff
@@ -0,0 +1,47 @@
...
+        uint32_t lo;
+    };
+};
+
+#define tt_l1_ptr __attribute__((rvtt_l1_ptr))
+#define tt_reg_ptr __attribute__((rvtt_reg_ptr))
+
+inline __attribute__((always_inline)) uint64_t tt_l1_load(tt_uint64_t tt_l1_ptr *p)
+{
+    tt_uint64_t v;
```
fvranicTT, line 23: The function separately loads hi and lo, but tt_uint64_t is structured as a union, meaning v already represents the combined value. Instead, all we need to do is `return p->v`.
Furthermore, I don't see this code being used anywhere. If that's the case, I'd propose removing it.

tests/firmware/riscv/common/risc_attribs.h
```diff
@@ -0,0 +1,47 @@
...
+
+    v.hi = p->hi;
+    v.lo = p->lo;
+    return v.v;
+}
+
+inline __attribute__((always_inline)) uint64_t tt_l1_load(volatile tt_uint64_t * tt_l1_ptr p)
+{
+    tt_uint64_t v;
```
fvranicTT, line 32: See comment above. Same remarks apply.

tests/firmware/riscv/common/tt_log.h
```diff
@@ -0,0 +1,147 @@
...
+  log_mailbox[0] = 0xC0FFEE;
+
+  // Wait for host to clear mailbox
+  if constexpr (Blocking) {
+    while (log_mailbox[0] != 0) {};
+  }
+}
+
+template <uint32_t Locator, bool Blocking, typename... Args>
+void ttlog(Args... args) {
```
fvranicTT, line 62: Seems to be unused, like the rest of the file.

tests/firmware/riscv/common/fw_debug.h
```diff
@@ -0,0 +1,8 @@
+// SPDX-FileCopyrightText: © 2023 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#define FWASSERT(s, p)
```
fvranicTT, line 5: The way it is defined FWASSERT doesn't do much except that it allows this macro to be called. It seems like we're missing the implementation?

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    if unpack_func:
+        return unpack_func(read_data_bytes)
+    else:
+        raise ValueError(f"Unsupported format: {format}")
+    
+def read_mailboxes():
+    mailbox_addresses = [0x19FF4, 0x19FF8, 0x19FFC]
+    mailbox_values = [read_words_from_device("0,0", address, word_count=1)[0].to_bytes(4, 'big') for address in mailbox_addresses]
+    return all(value == b'\x00\x00\x00\x01' for value in mailbox_values)
```
fvranicTT, line 89: I'd argue this is going to make debugging harder, since you won't know which mailbox isn't set to 1. So I'd assert them individually. Also, no need to convert value to bytes and compare with hex, comparing against == 1 seems enough, and it's easier to read.

tests/helpers/include/ckernel_helper.h
```diff
@@ -0,0 +1,24 @@
...
+	uint32_t math_sync_tile_dst_index = 0;
+	volatile uint tt_reg_ptr *mailbox_base[4] = {
+    reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX0_BASE), reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX1_BASE),
+    reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX2_BASE), reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX3_BASE)};
+
+	volatile uint32_t inst_trace_ptr  __attribute__((section(".init"))) = 0;
+	volatile uint32_t inst_trace[1024]  __attribute__((section(".init"))) = {0};
+
+	uint32_t cfg_state_id __attribute__((section(".bss"))) = 0;  // Flip between 0 and 1 to keep state between kernel calls
+	uint32_t dest_offset_id __attribute__((section(".bss"))) = 0; // Flip between 0 and 1 to keep dest pointer between kernel calls
```
fvranicTT, line 18:
```suggestion
	std::uint32_t cfg_state_id = 0;  // Flip between 0 and 1 to keep state between kernel calls
	std::uint32_t dest_offset_id = 0; // Flip between 0 and 1 to keep dest pointer between kernel calls
```
Global variables initialized to zero are placed in the .bss segment by default. Could you please check if removing the attribute changes the compilation outcome?

tests/helpers/include/ckernel_helper.h
```diff
@@ -0,0 +1,24 @@
...
+	volatile uint tt_reg_ptr *instrn_buffer = reinterpret_cast<volatile uint *>(INSTRN_BUF_BASE);
+	volatile uint tt_reg_ptr *regfile = reinterpret_cast<volatile uint *>(REGFILE_BASE);
+	volatile uint tt_l1_ptr * trisc_l1_mailbox = reinterpret_cast<volatile uint tt_l1_ptr *>(0x1d000);
+	uint32_t math_sync_tile_dst_index = 0;
+	volatile uint tt_reg_ptr *mailbox_base[4] = {
+    reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX0_BASE), reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX1_BASE),
+    reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX2_BASE), reinterpret_cast<volatile uint tt_reg_ptr *>(TENSIX_MAILBOX3_BASE)};
+
+	volatile uint32_t inst_trace_ptr  __attribute__((section(".init"))) = 0;
+	volatile uint32_t inst_trace[1024]  __attribute__((section(".init"))) = {0};
```
fvranicTT, line 15:
```suggestion
```
I don't see these two used anywhere in the tt_llk repo.

tests/helpers/include/params.h
```diff
@@ -0,0 +1,109 @@
+#ifndef PARAMS_H
+#define PARAMS_H
+
+#include <cstdint>
+#include <cstdarg>
+
+#ifdef LLK_TRISC_UNPACK
```
fvranicTT, line 7: Can we scope this with a namespace? Relying so heavily on preprocessor directives could prove to be error-prone. Also, C-style casting is not the safest way to manipulate data.

```cpp
template <typename DataFormat>
constexpr auto to_underlying(DataFormat e) noexcept {
    return static_cast<std::underlying_type_t<DataFormat>>(e);
}

#ifdef LLK_TRISC_UNPACK
    #ifdef FORMAT_FLOAT16_B
        #define DATA_FORMAT to_underlying(DataFormat::Float16_b)
    #elif defined(FORMAT_FLOAT16)
        #define DATA_FORMAT to_underlying(DataFormat::Float16)
    #elif defined(FORMAT_FLOAT32)
        #define DATA_FORMAT to_underlying(DataFormat::Float32)
    #elif defined(FORMAT_INT32)
        #define DATA_FORMAT to_underlying(DataFormat::Int32)
    #elif defined(FORMAT_BFP8_B)
        #define DATA_FORMAT to_underlying(DataFormat::Bfp8_b)
    #endif
#endif
```

tests/helpers/src/trisc.cc
```diff
@@ -0,0 +1,41 @@
...
+#include <l1_address_map.h>
+#include <tensix.h>
+// Necessary for ckernel variables
+#include "ckernel_helper.h"
+#include "params.h"
+
+#ifdef LLK_TRISC_UNPACK
+    volatile uint32_t* mailbox = (volatile uint32_t*)(0x19FFC);
+#elif defined(LLK_TRISC_MATH)
+	volatile uint32_t* mailbox = (volatile uint32_t*)(0x19FF8);
```
fvranicTT, line 19: space vs tab formatting

tests/helpers/src/trisc.cc
```diff
@@ -0,0 +1,41 @@
...
+#elif defined(LLK_TRISC_PACK)
+	volatile uint32_t* mailbox = (volatile uint32_t*)(0x19FF4);
+#endif
+
+
+int main()
+{
+
+	*mailbox = 0x2;
```
fvranicTT, line 28: can you add a comment to explain this magic number (0x2)?

tests/python_tests/test_eltwise_unary_sfpu.py
```diff
@@ -0,0 +1,97 @@
+import pytest
+import torch
+import os
+import math
+from helpers import *
+
+def generate_golden(operation, operand1, data_format):
```
fvranicTT, line 7: Would this work?
```python
def generate_golden(operation, operand1, data_format):
    tensor1_float = operand1.clone().detach().to(format_dict[data_format] if data_format != "Bfp8_b" else torch.bfloat16)

    if operation == "sqrt":
        res = torch.sqrt(tensor1_float)
    elif operation == "square":
        res = torch.square(tensor1_float)
    elif operation == "log":
        res = torch.log(tensor1_float)
        res[tensor1_float == 0] = float('nan')  # Handle log(0) case
    else:
        raise ValueError("Unsupported operation!")

    # Get the first face
    return res[:256].tolist()
```

tests/sources/unpack_tilize_test.cc
```diff
@@ -0,0 +1,101 @@
...
+// Globals
+uint32_t unp_cfg_context = 0;
+uint32_t pack_sync_tile_dst_ptr = 0;
+volatile uint32_t tt_l1_ptr l1_buffer[16] __attribute__ ((section (".text#"))) __attribute__ ((aligned (16)));
+
+#ifdef DEST_ACC
+const bool is_fp32_dest_acc_en = true;
+#else
+const bool is_fp32_dest_acc_en = false;
+#endif
```
fvranicTT, line 17:
```suggestion
const bool is_fp32_dest_acc_en = 
#ifdef DEST_ACC
    true;
#else
    false;
#endif
```

tests/sources/unpack_tilize_test.cc
```diff
@@ -0,0 +1,101 @@
...
+
+#endif 
+
+#ifdef LLK_TRISC_PACK
+
+#include "llk_pack.h"
+#include "llk_pack_common.h"
+#include "params.h"
+
+const bool UNTILIIZE = false;
```
fvranicTT, line 73: please put this variable inside run_kernel() method.

tests/sources/sfpu_binary_test.cc
```diff
@@ -0,0 +1,120 @@
...
+
+#endif 
+
+#ifdef LLK_TRISC_PACK
+
+#include "llk_pack.h"
+#include "llk_pack_common.h"
+#include "params.h"
+
+volatile uint32_t* buffer_Dest = (volatile uint32_t*)0x1c000;
```
fvranicTT, line 94: Can you move this variable to run_kernel()?

tests/helpers/src/brisc.cc
fvranicTT, line 1: Can we rename all .cc files to .cpp? Renaming the files to .cpp would align with the more commonly used file extension for C++ source files. This change can improve consistency and readability within the project, as .cpp is widely recognized and used in the C++ community. Additionally, it may help new contributors understand the project's structure more quickly.

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#ifdef ARCH_BLACKHOLE
+    WRITE_REG(RISCV_DEBUG_REG_NCRISC_RESET_PC, MEM_NCRISC_FIRMWARE_BASE);
+    WRITE_REG(RISCV_DEBUG_REG_TRISC0_RESET_PC, MEM_TRISC0_FIRMWARE_BASE);
+    WRITE_REG(RISCV_DEBUG_REG_TRISC1_RESET_PC, MEM_TRISC1_FIRMWARE_BASE);
+    WRITE_REG(RISCV_DEBUG_REG_TRISC2_RESET_PC, MEM_TRISC2_FIRMWARE_BASE);
+    WRITE_REG(RISCV_DEBUG_REG_TRISC_RESET_PC_OVERRIDE, 0b111);
+    WRITE_REG(RISCV_DEBUG_REG_NCRISC_RESET_PC_OVERRIDE, 0x1);
+#else
+    // cfg_regs[NCRISC_RESET_PC_PC_ADDR32] = MEM_NCRISC_FIRMWARE_BASE;
```
fvranicTT, line 33: Can we remove this part of the code, so we wouldn't be committing commented code?

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#ifdef ARCH_BLACKHOLE
+    *((uint32_t volatile*)RISCV_DEBUG_REG_DEST_CG_CTRL) = 0;
+#endif
+
+    #ifdef ARCH_BLACKHOLE // TODO see what happens with WH
+    WRITE_REG(RISCV_TDMA_REG_CLK_GATE_EN, 0x3f);  // Enable clock gating
+    set_deassert_addresses();
+    // Invalidate tensix icache for all 4 risc cores
+    cfg_regs[RISCV_IC_INVALIDATE_InvalidateAll_ADDR32] = RISCV_IC_BRISC_MASK | RISCV_IC_TRISC_ALL_MASK | RISCV_IC_NCRISC_MASK;
```
fvranicTT, line 60: Why not use WRITE_REG here as well?

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+    TTI_SEMINIT(1,0,ckernel::semaphore::UNPACK_TO_DEST);
+    TTI_SEMINIT(1,0,ckernel::semaphore::MATH_DONE);
+}
+
+void device_setup() {
+
+    volatile tt_reg_ptr uint32_t* cfg_regs = reinterpret_cast<uint32_t*>(TENSIX_CFG_BASE);
+
+#ifdef ARCH_BLACKHOLE
+    *((uint32_t volatile*)RISCV_DEBUG_REG_DEST_CG_CTRL) = 0;
```
fvranicTT, line 53: Can we `WRITE_REG(RISCV_DEBUG_REG_DEST_CG_CTRL, 0x0);` instead? Looks easier to understand.
Also, it'd be nice if you could add a comment about the outcome of this command.

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+#endif
+
+    #ifdef ARCH_BLACKHOLE // TODO see what happens with WH
```
fvranicTT, line 56: You could merge these ifdefs, since they are both for BH.
```cpp
#ifdef ARCH_BLACKHOLE
WRITE_REG(RISCV_DEBUG_REG_DEST_CG_CTRL, 0x0);
WRITE_REG(RISCV_TDMA_REG_CLK_GATE_EN, 0x3f);  // Enable clock gating
set_deassert_addresses();
// Invalidate tensix icache for all 4 risc cores
WRITE_REG(RISCV_IC_INVALIDATE_InvalidateAll_ADDR32, RISCV_IC_BRISC_MASK | RISCV_IC_TRISC_ALL_MASK | RISCV_IC_NCRISC_MASK);
TTI_ZEROACC(p_zeroacc::CLR_ALL, 0, 0, 1, 0);
#else
TTI_ZEROACC(p_zeroacc::CLR_ALL, 0, 0);
#endif
```

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+	TTI_SFPLOADI(p_sfpu::LREG0,0xA,0xbf80); // -1.0f -> LREG0
+	TTI_SFPCONFIG(0, 11, 0); // LREG0 -> LREG11
+
+    initialize_tensix_semaphores();
+
+}
+
+int main(){
+    device_setup();
+    for(;;){}
```
fvranicTT, line 84: I know you've mentioned something about issues when using while loop here. Maybe it got optimized away?
This should do the trick:
```cpp
int main() {
    device_setup();
    // Use a volatile variable to prevent the compiler from optimizing away the loop
    volatile bool run = true;
    
    // Infinite loop
    while (run) { }
}
```

tests/helpers/src/trisc.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "ckernel_main.h"
+#include "ckernel_globals.h"
+#include <l1_address_map.h>
+#include <tensix.h>
+// Necessary for ckernel variables
+#include "ckernel_helper.h"
+#include "params.h"
+
+#ifdef LLK_TRISC_UNPACK
+	volatile uint32_t* mailbox = (volatile uint32_t*)(0x19FFC);
```
fvranicTT, line 17: Is there a predefined value for these mailboxes (e.g. `TRISC0_MAILBOX_BASE, TRISC1_MAILBOX_BASE, TRISC2_MAILBOX_BASE`, in that order)? Something from the BH/WH header files you pulled in from tt-metal that defines these? I'm looking into making this future-proof, so that if the address changes these tests continue to operate.
This variable could be defined inside main vs globally, correct?

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,89 @@
+from ttlens.tt_lens_init import init_ttlens
+from ttlens.tt_lens_lib import write_to_device, read_words_from_device, run_elf
+from helpers import *
+
+ELF_LOCATION = "../build/elf/"
```
fvranicTT, line 5: Please move this variable into `run_elf_files` method.

tests/helpers/src/trisc.cpp
```diff
@@ -0,0 +1,41 @@
...
+#endif
+
+
+int main()
+{
+
+	*mailbox = 0x2; // write value different than 1 to mailbox to indicate kernel is running
+
+    //FWEVENT("Launching proudction env kernels");
+	for (int i = 0; i < 64; i++) regfile[i] = 0;
```
fvranicTT, line 31: memset would maybe make it more evident what you're trying to do. Missing { } got me for a second.

tests/helpers/tmu-crt0.S
```diff
@@ -0,0 +1,159 @@
+.section .init
```
tt-vjovanovic, line 1: For this file and ld files that are copied from tt-lens, we will need to figure out what will be the best way to share them. Considering external users, it is not the best way to copy them (in case we change something, or you want to change something, etc).
We will come back to this once we implement full debugging experience...

tests/helpers/src/brisc.cpp
```diff
@@ -0,0 +1,85 @@
...
+inline void WRITE_REG(uint32_t addr, uint32_t val) {
+    volatile tt_reg_ptr uint32_t* ptr = reinterpret_cast<uint32_t*>(addr);
+    ptr[0] = val;
+}
+
+inline void set_deassert_addresses() {
+    volatile tt_reg_ptr uint32_t* cfg_regs = reinterpret_cast<uint32_t*>(TENSIX_CFG_BASE);
+
+#ifdef ARCH_BLACKHOLE
+    WRITE_REG(RISCV_DEBUG_REG_NCRISC_RESET_PC, MEM_NCRISC_FIRMWARE_BASE);
```
tt-vjovanovic, line 26: Note that `load_elf` is also setting these registers and if you run brisc at the same time as other elfs it will override it and it won't work as expected....

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,89 @@
...
+
+def collect_results(format,address=0x1c000,sfpu=False):
+    read_words_cnt = calculate_read_words_count(format,sfpu)
+    read_data = read_words_from_device("0,0", address, word_count=read_words_cnt)
```
tt-vjovanovic, line 8: Consider using `"0,0"` as a function argument... Maybe in the future you have multiple cores that you want to run...

tests/python_tests/helpers/pack.py
```diff
@@ -0,0 +1,111 @@
+# pack.py
+
+import struct
+import torch
+import struct
+
+def flatten_list(sublists):
+    return [item for sublist in sublists for item in sublist]
+
+def int_to_bytes_list(n):
```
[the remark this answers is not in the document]
tt-vjovanovic: or:
```
def int_to_bytes_list(n):
    return [(n >> i) & 0xFF for i in range(0, 32, 8)]
```

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,89 @@
...
+        unpack_func = unpackers.get(format)
+    
+    if unpack_func:
+        return unpack_func(read_data_bytes)
+    else:
+        raise ValueError(f"Unsupported format: {format}")
+    
+def read_mailboxes():
+    mailbox_addresses = [0x19FF4, 0x19FF8, 0x19FFC]
+    mailbox_values = [read_words_from_device("0,0", address, word_count=1)[0].to_bytes(4, 'big') for address in mailbox_addresses]
```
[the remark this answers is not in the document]
fvranicTT: Yes, but ideally, we'd like to assert them one by one, to know which one isn't 1, if that situation happens. As for the reading part, I agree, no need to covert to bytes. :)

tests/python_tests/helpers/pack.py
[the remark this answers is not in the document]
fvranicTT: Even cleaner 👍

tests/python_tests/test_sfpu_binary.py
```diff
@@ -0,0 +1,91 @@
...
+    run_elf_files(testname)
+    
+    res_from_L1 = collect_results(format)
+
+    assert len(res_from_L1) == len(golden)
+
+    run_shell_command("cd .. && make clean")
+
+    # Mailbox checks
+    assert read_words_from_device("0,0", 0x19FF4, word_count=1)[0].to_bytes(4, 'big') == b'\x00\x00\x00\x01'
```
tt-vjovanovic, line 73: Reuse existing function that reads all mailboxes?

tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -0,0 +1,71 @@
...
+    make_cmd = generate_make_command(test_config)
+    run_shell_command(f"cd .. && {make_cmd}")
+    run_elf_files(testname)
+
+    res_from_L1 = collect_results(format)
+
+    run_shell_command("cd .. && make clean")
+
+    assert len(res_from_L1) == len(golden)
+    assert read_mailboxes() == True
```
fvranicTT, line 55: What if it isn't (True)? Which one caused the failure? I think it would be easier to either have that in the tests, or to assert them separately in a function (just name it more explicitly, so that everyone knows what it does, before looking into it).

tests/python_tests/test_multiple_tiles_eltwise.py
```diff
@@ -0,0 +1,128 @@
...
+                element = element.to(torch.int32)  # Convert to int32
+                element &= 0xFFFE
+        elif math_fidelity == 2:  # HiFi2
+            for element in operand2:
+                element = element.to(torch.int32)  # Convert to int32
+                element &= 0xFFFE
+        elif math_fidelity == 3:  # HiFi3
+            pass
+        elif math_fidelity == 4:  # HiFi4
+            pass
```
fvranicTT, line 38:
```suggestion
    if data_format == "Float16_b":
        if math_fidelity in [0, 2]:  # LoFi or HiFi2
            for element in operand2:
                element = element.to(torch.int32)
                element &= 0xFFFE
        if math_fidelity == 0:  # LoFi
            for element in operand1:
                element = element.to(torch.int32)
                element &= 0xFFF8
```

tests/python_tests/helpers/device.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    if unpack_func:
+        return unpack_func(read_data_bytes)
+    else:
+        raise ValueError(f"Unsupported format: {format}")
+    
+def read_mailboxes(core_loc : str= "0,0"):
+    mailbox_addresses = [0x19FF4, 0x19FF8, 0x19FFC] # L1 Mailbox addresses
+    mailbox_values = [read_word_from_device(core_loc, address) for address in mailbox_addresses]
+    return all(value == 1 for value in mailbox_values)
```
fvranicTT, line 89:
```suggestion
def assert_tensix_operations_finished(core_loc : str= "0,0"):
    mailbox_addresses = [0x19FF4, 0x19FF8, 0x19FFC] # L1 Mailbox addresses
    assert read_word_from_device(core_loc, tensix_mailboxes[0]) == 1
    assert read_word_from_device(core_loc, tensix_mailboxes[1]) == 1
    assert read_word_from_device(core_loc, tensix_mailboxes[2]) == 1
```

tests/python_tests/helpers/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .stimuli_generator import flatten_list,generate_stimuli
+from .format_arg_mapping import format_dict,format_args_dict,mathop_args_dict,format_sizes
+from .pack import pack_bfp16,pack_fp16,pack_fp32,pack_int32,pack_bfp8_b
+from .unpack import unpack_fp16,unpack_bfp16,unpack_float32,unpack_int32,unpack_bfp8_b, int_to_bytes_list
+from .utils import run_shell_command,calculate_read_words_count,tilize,untilize,compare_pcc, format_kernel_list, print_faces
+from .device import collect_results, run_elf_files, write_stimuli_to_l1, get_result_from_device, assert_tensix_operations_finished
+from .test_config import generate_make_command
+from ttlens import Verbosity
```
tt-vjovanovic, line 8: Note that after https://github.com/tenstorrent/tt-exalens/pull/267 new name will be `ttexalens`

tests/helpers/include/ckernel_helper.h
```diff
@@ -0,0 +1,21 @@
+#ifndef CKERNEL_HELPER_H
+#define CKERNEL_HELPER_H
+
+namespace ckernel{
+	volatile std::uint32_t tt_reg_ptr *pc_buf_base = reinterpret_cast<volatile std::uint32_t *>(PC_BUF_BASE);
```
fvranicTT, line 9: Please add "risc_attrbs.h", as the definition of `tt_reg_ptr` is located there. Also, add `<cstdint>` header, too.

PR #18, opened 2025-02-24
build: introduce pre-commit

fvranicTT:
**Description:**
This PR introduces consistent formatting for all C++ and Python files to ensure uniform code style across the repository. While this is not a final set of formatting rules, it serves as an initial baseline that can be refined over time.

Since this change affects most files, merging it will help keep future PRs clean and focused on functional changes rather than formatting inconsistencies. After merging, I will follow up with another PR to remove this change from git history, preventing unnecessary git blame noise.

**Next Steps:**
- Specific formatting rules can be adjusted as needed.
- A follow-up PR will clean up the commit history to maintain meaningful git blame tracking.

Let me know if you have any concerns or suggestions!


vmilosevic (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  rev: v1.0.3
+  hooks:
+    - id: check-copyright
+      args: ['--config', 'infra/check_copyright_config.yaml', "--ignore", ""]
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v3.4.0
+  hooks:
+    - id: trailing-whitespace
+    - id: end-of-file-fixer
+    - id: check-added-large-files
```
ncvetkovicTT, line 27: Is the definition of "large" flexible? Don't know why I would want to add large files at all, just curious.

ncvetkovicTT:
@fvranicTT Do we maybe want to block binaries from being committed as well? Are they blocked by default somehow? If so, is the filter configurable, for instance if we want to add documentation in the future and some .png sneaks in as a screenshot instead of a .svg.

ncvetkovicTT (approved):
Thanks

.pre-commit-config.yaml
fvranicTT: By default, this is 500 KB. I will set it explicitly, so it's better described and understood.

fvranicTT:
> @fvranicTT Do we maybe want to block binaries from being committed as well? Are they blocked by default somehow? If so, is the filter configurable, for instance if we want to add documentation in the future and some .png sneaks in as a screenshot instead of a .svg.

A very good question. I tried blocking binaries with `check-added-large-files`, which by default blocks everything above 500 KB. We can set this limit higher, and PNG images are good example why we might want to have these files. pre-commit can't block particular extensions, e.g. so we could forbid commiting .so files by mistake.

PR #26, opened 2025-02-26
#18064: Guard CSRRS by disabling branch prediction

ncvetkovicTT:
Due to a HW bug in CSR writes, we need to guard them by first disabling branch prediction, and then re-enabling it. This is a workaround for https://github.com/tenstorrent/tt-metal/issues/18064#issuecomment-2685621251 and is proposed as a fix by the HW team.


ncvetkovicTT:
@ttmtrajkovic @amahmudTT If this looks right to you, let's unblock metal.

ttmtrajkovic (approved):

PR #29, opened 2025-02-27
fix: include missing headers and add missing files

fvranicTT:
## Description

This PR adds missing headers and hardware-specific headers for both blackhole and wormhole, so tt-llk could build in tt-metal even after formatting and sorting the header files. At the moment, that's not the case, and current setup works only because of circumstances.

How did we spot the problem? After applying clang-format and sorting the header files, llk started failing during compilation, due to unrecognized tokens.

What was added: 
 1.  The hw_specific directory contains four essential files: cfg_defines.h, tensix_types.h, tensix.h, and dev_mem_map.h. These files are hardware-specific and include important definitions, global variables, and methods used in the llk code. However, they are currently not included in the build process, which results in tt-llk being unbuildable in standalone variant.
 2. missing includes (all sorts of them).

Checks:
- [x] Blackhole [post-commit checks](https://github.com/tenstorrent/tt-metal/actions/runs/13585628575)
- [x] All [post-commit checks](https://github.com/tenstorrent/tt-metal/actions/runs/13585626149)


ldjurovicTT (approved):

ttmtrajkovic:
@fvranicTT,

To confirm, when doing build in tt-metal env, the files from `hw_specifc` folder wouldnt get included/used?
Those files would only be used when tt_llk is compiled standalone, without tt-metal present?

Milos

fvranicTT:
@ttmtrajkovic That's the idea, yes.

PR #33, opened 2025-03-03
fix: disable failing python tests

fvranicTT:
## Summary
As we set up new test infrastructure, we've encountered several tests that are currently failing. This PR disables these failing tests so that test runs only include those that are known to be correct and passing.

## Motivation
Running incomplete or misconfigured failing tests can create unnecessary confusion, especially for developers making unrelated changes. They might mistakenly believe their modifications introduced new failures, making debugging more difficult. By disabling these tests, we ensure that test failures are meaningful and actionable.

## Next Steps
Once the new test infrastructure is fully set up, we should revisit these disabled tests and re-enable them after proper fixes or updates.
Future work may involve categorizing or marking these tests separately for better tracking.

### DIFF
--- a/tests/python_tests/test_eltwise_unary_datacopy.py
+++ b/tests/python_tests/test_eltwise_unary_datacopy.py
@@ -29,13 +29,16 @@ param_ids = [
 
 def test_unary_datacopy(format, testname, dest_acc):
 
+    if (format == "Float16" and dest_acc == "DEST_ACC"):
+        pytest.skip(reason = "This combination is not fully implemented in testing")
+
+    if( format in ["Float32", "Int32"] and dest_acc!="DEST_ACC"):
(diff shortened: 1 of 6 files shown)


tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -29,14 +29,17 @@ def generate_golden(operand1,format):
 
 def test_unary_datacopy(format, testname, dest_acc):
 
+    if (format == "Float16" and dest_acc == "DEST_ACC"):
+        pytest.skip(reason = "This combination is not fully implemented in testing")
+
+    if( format in ["Float32", "Int32"] and dest_acc!="DEST_ACC"):
```
fvranicTT, line 35:
```suggestion
    if(format in ["Float32", "Int32"] and dest_acc!="DEST_ACC"):
```

PR #41, opened 2025-03-05
Fix: Use TT_* instruction type instead of TTI_* as the parameter is not compile time known

nvelickovicTT:
General rule is:

TT_* type of instruction is used when not all arguments are known at compile time.
TTI_* type of instruction is used when all argument are compile time constants.

Former is slower and more general, latter results in more optimal and faster code.

Note: Could be that using non const (non constexpr) variable in TTI_* type of instruction works in cases when compiler deduces that some variable depends only on compile time known values. But, this is unsafe to rely on.


ncvetkovicTT (approved):
lgtm!

PR #44, opened 2025-03-07
Added narrow Row LLK support

pmilenkovicTT:
The issue was that the narrow row feature did not exist on the BH board.
These additions modify the LLK to accept narrow_row as a parameter and change the algorithm to use 1 packer interface if the data arrives in a narrow row manner (narrow row relies on only having Face 0 and Face 2 in the tile, so we have to skip the Face 1/Face 3 reads).

The modifications were tested with the expected input and provide the same output as the WH_B0.


rtawfik01 (approved):

fvranicTT:
Can we have some kind of test on tt-metal that confirms this feature is working as expected, and prevents future regression?

PR #46, opened 2025-03-10
[Blackhole] Fix undefined numbers appearing in div SFPU OP

nvelickovicTT:
Div OP, which uses reciprocal underneath, produces unexpected numbers on Blackhole in specific situations.

Issue and fix reported in https://github.com/tenstorrent/tt-metal/issues/18496.
Investigation and fix provided by @umadevimcw in the original issue.

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_recip.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_recip.h
@@ -69,7 +69,14 @@ inline void _calculate_reciprocal_(const int iterations)
         }
         v_endif;
 
-        dst_reg[0] = reinterpret<vFloat>(float_to_fp16b(out, 0));
+        if constexpr (APPROXIMATION_MODE)
+        {
+            dst_reg[0] = out;
+        }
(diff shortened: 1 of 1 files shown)


lpremovicTT (approved):
While these changes make sense and are ok to merge, i would like to raise two points for later investigation.

- Is there an explicit guarantee that APPROXIMATION_MODE true will never generate an output that doesnt fit into fp16b, if not, should both versions perform float to fp16b conversion.
- Where this if makes more sense is in WH where recip has is_fp32_dest_acc_en template parameter and if dest is 32 naturally we wouldn't want to convert float to fp16b, but this parameter is absent on BH impl and also WH recip is the only occurrence of this parameter in both WH and BH. Should this parameter be added somewhere else or removed or is there a good reason for it only being used there.

lpremovicTT:
For BH I think the more sane version would be to always convert, we should test that when we have the time.

nvelickovicTT:
Feedback from @rdjogoTT who implemented this optimization for WH: No need to spend cycles on rounding when in `APPROXIMATION_MODE` mode.

PR #54, opened 2025-03-13
test: reset card after each test run

fvranicTT:
### Ticket
None

### Problem description
We have observed a recurring issue where tests begin to fail consecutively, one after another. This pattern suggests that the state of the chip after the completion of one test is somehow corrupted. Potential issues include the chip being left in an illegal state, memory not being cleared, or other side effects that interfere with subsequent tests. These residual effects from previous tests can cause subsequent tests to fail, even if they would otherwise pass under normal conditions. This situation undermines the reliability and accuracy of our test results, making it difficult to identify genuine issues in the code.

### What's changed
This change introduces a card reset call during the teardown phase of each Python test. By doing so, it ensures that the hardware is reset to a clean state between test runs. This is particularly important when running a large number of tests, as it prevents potential side effects from previous tests affecting subsequent ones. 

### Type of change
- [ ] Documentation update

### DIFF
--- a/tests/helpers/src/trisc.cpp
+++ b/tests/helpers/src/trisc.cpp
@@ -18,13 +18,13 @@ int main()
 {
     volatile std::uint64_t* TIMESTAMP_ADDRESS = reinterpret_cast<volatile std::uint64_t*>(0x19000);
 #if defined(LLK_TRISC_UNPACK)
-    static constexpr uint32_t core_idx = 0;
+    const std::uint32_t core_idx       = 0;
     volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FFC);
 #elif defined(LLK_TRISC_MATH)
-    static constexpr uint32_t core_idx = 1;
+    const std::uint32_t core_idx       = 1;
     volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FF8);
 #elif defined(LLK_TRISC_PACK)
-    static constexpr uint32_t core_idx = 2;
+    const std::uint32_t core_idx       = 2;
     volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FF4);
 #endif
     std::uint64_t wall_clock = ckernel::read_wall_clock();
--- a/tests/helpers/src/trisc.cpp
+++ b/tests/helpers/src/trisc.cpp
@@ -27,7 +27,7 @@ int main()
     static constexpr uint32_t core_idx = 2;
     volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FF4);
 #endif
-    uint64_t wall_clock = ckernel::read_wall_clock();
+    std::uint64_t wall_clock = ckernel::read_wall_clock();
 
     *(TIMESTAMP_ADDRESS + core_idx * 2) = wall_clock;
     *mailbox                            = 0x2; // write value different than 1 to mailbox to indicate kernel is running
--- a/tests/helpers/src/trisc.cpp
+++ b/tests/helpers/src/trisc.cpp
@@ -16,23 +16,31 @@
 
 int main()
 {
-#ifdef LLK_TRISC_UNPACK
-    volatile std::uint32_t* mailbox = reinterpret_cast<volatile std::uint32_t*>(0x19FFC);
+    volatile std::uint64_t* TIMESTAMP_ADDRESS = reinterpret_cast<volatile std::uint64_t*>(0x19000);
+#if defined(LLK_TRISC_UNPACK)
+    static constexpr uint32_t core_idx = 0;
+    volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FFC);
 #elif defined(LLK_TRISC_MATH)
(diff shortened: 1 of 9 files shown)


tests/python_tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from helpers import HardwareController
+
+
+@pytest.fixture(scope="function", autouse=True)
```
fvranicTT, line 8: Note to reviewers: when `autouse` is set to True, this means that all tests located in the same directory as `conftest.py` should use this fixture during test setup and teardown (we're interested in this second thing). So effectively, there's no need to update all other tests by passing this method in test definition.

skotaracTT: This is great! Keeps tests calls clear and less clutter.

skotaracTT:
Looked for a mistake... didn't find one 😎 
everything is clear and minimal changes required. Great

skotaracTT (approved):

tests/python_tests/helpers/hardware_controller.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self):
+        self.chip_architecture = get_chip_architecture()
+
+    def reset_card(self):
+        if self.chip_architecture == "blackhole":
+            print("Resetting BH card")
+            run_shell_command("/home/software/syseng/bh/tt-smi -r 0")
+        elif self.chip_architecture == "wormhole":
+            print("Resetting WH card")
+            run_shell_command("/home/software/syseng/wh/tt-smi -wr 0")
```
nvelickovicTT, line 21: JFYI, this path is not available on aus cluster.

nvelickovicTT: Also, this path is provided as a convenience, but with no guarantees. The inteded way of consuming tt-smi is by pulling in a release from GH, or clone and build.

fvranicTT: OK, that means we need to use relative path and rely on our setup script for now. Not a big change. I was trying to use more "stable" way. Boy was I wrong!
Thanks for spotting this.

tests/python_tests/helpers/hardware_controller.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    def __init__(self):
+        self.chip_architecture = get_chip_architecture()
+
+    def reset_card(self):
+        if self.chip_architecture == "blackhole":
+            print("Resetting BH card")
+            run_shell_command("tt-smi -r 0")
+            time.sleep(1)
```
fvranicTT, line 20:
```suggestion
```

tests/python_tests/helpers/device.py
```diff
@@ -117,7 +123,7 @@ def assert_value_with_timeout(core_loc, mailbox_addr, timeout=0, poll_interval=0
 
     while time.time() < end_time:
         if read_word_from_device(core_loc, mailbox_addr) == 1:
-            return True
+            break
```
fvranicTT, line 126:
```suggestion
            return True
```

tests/python_tests/helpers/device.py
```diff
@@ -27,21 +30,27 @@ def collect_results(
...
     ELF_LOCATION = "../build/elf/"
 
     if run_brisc:
         run_elf(f"{ELF_LOCATION}brisc.elf", core_loc, risc_id=0)
 
-    # for i in range(3):
-    #     run_elf(f"{ELF_LOCATION}{testname}_trisc{i}.elf", core_loc, risc_id=i + 1)
+    context = check_context()
+    device = context.devices[0]
+    RISC_DBG_SOFT_RESET0 = device.get_tensix_register_address("RISCV_DEBUG_REG_SOFT_RESET_0")
```
fvranicTT, line 40:
```suggestion
    RISC_DBG_SOFT_RESET0 = device.get_tensix_register_address(
            "RISCV_DEBUG_REG_SOFT_RESET_0"
    )
```

tests/python_tests/helpers/hardware_controller.py
fvranicTT: updated to use tt-smi from PATH.

tests/helpers/src/trisc.cpp
```diff
@@ -16,23 +16,31 @@
...
-    volatile std::uint32_t* mailbox = reinterpret_cast<volatile std::uint32_t*>(0x19FF8);
+    static constexpr uint32_t core_idx = 1;
+    volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FF8);
 #elif defined(LLK_TRISC_PACK)
-    volatile std::uint32_t* mailbox = reinterpret_cast<volatile std::uint32_t*>(0x19FF4);
+    static constexpr uint32_t core_idx = 2;
+    volatile std::uint32_t* mailbox    = reinterpret_cast<volatile std::uint32_t*>(0x19FF4);
 #endif
-    *mailbox = 0x2; // write value different than 1 to mailbox to indicate kernel is running
+    uint64_t wall_clock = ckernel::read_wall_clock();
```
fvranicTT, line 30:
```suggestion
    std::uint64_t wall_clock = ckernel::read_wall_clock();
```

PR #43, opened 2025-03-06
Add PRNG to BH dropout

JushBJJ:
**Ticket**
https://github.com/tenstorrent/tt-metal/issues/7736

This PR copies Wormhole's PRNG for dropout into blackhole

**Notes**
First, test to see if `ckernel::wait` works on blackhole, otherwise fallback to the commented code inside `init_prng_seed`. See https://github.com/tenstorrent/tt-metal/issues/10868

Second, I've created a tt-metal branch with the complete changes for the dropout ckernel headers that are in `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu`, and routes the tt_llk submodule to this PR's branch. 

You can use that fork to test as if the full change is applied for both repos:
https://github.com/JushBJJ/tt-metal/tree/bh-dropout

PR #58, opened 2025-03-17
fix: use ckernel::wait() instead of issuing NOPs during prng init

fvranicTT:
### Problem description
During the review of https://github.com/tenstorrent/tt-llk/pull/43, it was observed that NOPs were being used instead of ckernel::wait() to wait for the PRNG to accumulate data. Measuring the time it took to generate 600 NOPs revealed that it introduced more delay than initially planned. According to Syed, around 500 cycles are needed to acquire enough random data, but the NOP-based wait introduced a delay of 2200 cycles.

### What's changed
This PR updates the `ckernel::wait()` method to use `ckernel::read_wall_clock()` for calculating wait cycles more accurately. Since the method is inlined, there should be minimal overhead. Testing at the tt-metal level showed that this approach allows for more granular control of delays, with `ckernel::wait(500)` producing delays of approximately 520 cycles.

### Type of change

### Checklist


tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -832,12 +829,8 @@ inline void init_prng_seed(const uint seed)
     // The seed for PRNG should at least be initialzied during chip bootup time.
     volatile uint tt_reg_ptr *cfg  = get_cfg_pointer();
     cfg[PRNG_SEED_Seed_Val_ADDR32] = seed;
-
-    // TODO: ckernel::wait does not work properly. Use ckernel::wait when fixed.
-    for (int i = 0; i < 600; i++)
-    {
-        TTI_SFPNOP;
-    }
+    // PRNG needs some 500 cycles to start generating values (source: Syed)
```
amahmudTT, line 832: Not sure if it is a good or a bad idea to put names in comments. Although this info might be helpful.

amahmudTT: to start generating values -> to initialaize 

It can generate without waiting, it is just that the generated values will be done using some old seed or something.

fvranicTT: Thanks for the clarification. I removed Syed's name (now it's HW design team) and amended the comment to state "to initialize". Thanks Anil!

tt_llk_wormhole_b0/common/inc/ckernel.h
```diff
@@ -617,8 +617,10 @@ inline void init_prng_seed(const uint seed)
     volatile uint tt_reg_ptr *cfg  = get_cfg_pointer();
     cfg[PRNG_SEED_Seed_Val_ADDR32] = seed;
 
+    // PRNG needs some 500 cycles to initialize (source: HW design team)
     // TODO: ckernel::wait does not work properly. Use ckernel::wait when fixed.
-    for (int i = 0; i < 600; i++)
+    // 200 SFPNOPs take around 610 cycles
+    for (int i = 0; i < 200; i++)
```
fvranicTT, line 623: Had to revert to the old way. wait definitely doesn't work on WH, while it works on BH. FYI @amahmudTT

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -832,12 +829,8 @@ inline void init_prng_seed(const uint seed)
...
     volatile uint tt_reg_ptr *cfg  = get_cfg_pointer();
     cfg[PRNG_SEED_Seed_Val_ADDR32] = seed;
-
-    // TODO: ckernel::wait does not work properly. Use ckernel::wait when fixed.
-    for (int i = 0; i < 600; i++)
-    {
-        TTI_SFPNOP;
-    }
+    // PRNG needs some 500 cycles to initialize (source: HW design team)
+    wait(500);
```
nvelickovicTT, line 833: What happens if we use the PRNG without being initialized? Is it just less random, or causes a hang?
If it is the latter, I would suggest to add back the margin of 100 cycles that was targeted before.

tt_llk_wormhole_b0/common/inc/ckernel.h
nvelickovicTT: Spec lists SFPNOP as a single cycle instruction. I presume you got the 610 number by measuring experimentally. In case this is due to the overhead of looping caused by current compiler settings, it could stop working if compiler settings change in the future.
Have you tested also with the loop unrolled?

Issue #14, opened 2025-02-20
[Test Infra] Check golden generation for matmul with L1 accum


vmilicevicTT:

ldjurovicTT:
Just s o It is written somewhere I get only first two numbers bad on WH for matmul. Hope to fix it soon. And have basic single tile matmul running on both WH and BH

vmilicevicTT:
@skotaracTT  to confirm this was completed, and close the issue

skotaracTT:
Golden generation is made for matmul tests. There is no existing matmul test for L1 accumulation.

PR #67, opened 2025-03-25
Enabled matmul test with 2 random tiles in Float16_b

ldjurovicTT:
### Ticket
[link](https://github.com/tenstorrent/tt-llk/issues/14)

### Problem description
Before this PR we had a problem with running matmul test with 2 completely random tiles. 

### What's changed
First change is the matmul test itself. Input matrices are stored as tilized in L1 with write_stimuli_to_l1. Torch's matmul was replaces by standard 3 for loop matrix multiply algorithm. Also cpp test was changed since it had a wrong template argument in math_hw_config. Packer code is cleaned up.
A bug was noted in unpack.py in python_tests/helpers where elements were being read in wrong order, since from all tests we run matmul shows that problem most clearly it was fixed. All other tests are not effected by this change. ( Please run them all before approving to find any sneaky bugs)

This applies only to BH. On WH test seems to give valid output but only first two numbers are not changed and reamin 0xdeadbeef. 

### Type of change

- [ ] Documentation update

### Checklist


tests/python_tests/helpers/pack.py
```diff
@@ -29,7 +29,7 @@ def bfloat16_to_bytes(number):
     for i in range(0, len(torch_tensor), 2):
         half1 = bfloat16_to_bytes(torch_tensor[i])
         half2 = bfloat16_to_bytes(torch_tensor[i + 1])
-        packed_bytes.extend([half1[0:2][::-1], half2[0:2][::-1]][::-1])
+        packed_bytes.extend([half1[0:2][::-1], half2[0:2][::-1]])
```
fvranicTT, line 32: Good catch!

tests/python_tests/test_matmul.py
```diff
@@ -18,13 +18,20 @@ def generate_golden(operand1, operand2, data_format, math_fidelity):
...
                 element &= 0xFFF8
 
-    return torch.matmul(
-        tilize(operand1).view(32, 32), tilize(operand2).view(32, 32)
-    ).view(-1)
+    operand1_matrix = operand1.view(32, 32).to(format_dict[data_format])
+    operand2_matrix = operand2.view(32, 32).to(format_dict[data_format])
+
+    result_matrix = torch.zeros(32, 32, dtype=operand1_matrix.dtype)
+    for i in range(32):
```
fvranicTT, line 25: Can this triple for-loop be replaced with `torch.matmul`?

tests/python_tests/test_matmul.py
```diff
@@ -86,6 +92,16 @@ def test_matmul(testname, formats, dest_acc, math_fidelity):
         ),
     )
 
+    print("*" * 50)
```
fvranicTT, line 95: This was handy for debugging I guess, but it shouldn't be in the production code.

tests/sources/matmul_test.cpp
```diff
@@ -55,7 +53,7 @@ void run_kernel()
 {
     _llk_math_matmul_init_<MATH_FIDELITY, DstTileFaceLayout::RowMajor>();
     _llk_math_pack_sync_init_<DstSync::SyncFull, is_fp32_dest_acc_en>();
-    _llk_math_hw_configure_<true, false>(MATH_FORMAT, MATH_FORMAT);
+    _llk_math_hw_configure_<false, false>(MATH_FORMAT, MATH_FORMAT);
```
fvranicTT, line 56: Can we put false, false into some variables?

tests/python_tests/helpers/pack.py
skotaracTT, line 32: Although this change fixes matmul test cases and is compatible with other tests in the infrastructure, we are currently only running tests where all data in pipeline is the same format. Bugs were previously found in `pack.py` and `unpack.py` when combining formats that were not the same. This change may override and reverse those bug patchups causing the same tests to start failing when we run them with various format combinations. This fix here is a patch up currently since we haven't found the universal fix for all tests across format combinations. Until we do a full format sweep I would suggest keeping this here but adding a flag that checks if input and output formats are the same, to avoid reversing order of bytes that works for differing format combinations.
```suggestion
        if formats.unpack_A_src == formats.pack_dst:
               packed_bytes.extend([half1[0:2][::-1], half2[0:2][::-1]])
```
  make sure to update function parameters 
`def pack_bfp16(torch_tensor, formats: FormatConfig):`

tests/python_tests/helpers/unpack.py
```diff
@@ -48,6 +48,9 @@ def bytes_to_bfloat16(byte_list):
         for i in range(0, len(limited_packed_list), 2)
     ]
 
+    for i in range(0, len(result), 2):
+        result[i], result[i + 1] = result[i + 1], result[i]
+
```
skotaracTT, line 53:
```suggestion
    if unpack_src == pack_dst:
        for i in range(0, len(result), 2):
            result[i], result[i + 1] = result[i + 1], result[i]

```
Similar to suggestion for `pack.py` you need to add condition in order to prevent reversing bug fix for format sweeps. On line 59 you will see that the same thing is done, consecutive byte pairs are reversed. This was a bug found in `unpack.py` when input format is `Bfp8_b` and the input and output formats are not the same. I suggest setting another flag that checks if our input and output is the same, otherwise when we run this test with a various format combinations the condition on line 59 will be true in some cases, which aims to reverse consecutive byte pairs, however this will already be done by loop at line 51 and will reverse the bug fix, causing the passing test to start failing.

tests/python_tests/test_matmul.py
ldjurovicTT: Probably, but I am not quite sure how it's implemented...

tests/sources/matmul_test.cpp
ldjurovicTT: I think there is no need for that since those false statements are for untilize_en and row_pool flags... Since this is matmul i think it is ok to leave it as false false

tests/python_tests/helpers/pack.py
ldjurovicTT: Can we run multiple combinations to see if this is really a fix or not ?

tests/python_tests/test_matmul.py
ldjurovicTT: It can be exchanged. I checked now

tests/python_tests/helpers/pack.py
ldjurovicTT: Thanks! But we should sit down and rethink all this stuff. Most of these functions need to be rewritten I think. And once We do it in much simpler way this kind of errors will never occur again

fvranicTT (approved):

PR #68, opened 2025-03-25
[Test Infrastructure] Fixed bug in test_tilize_calculate_untilize_L1 

skotaracTT:
### Problem description
Once TRISCs were synced to run at the same time in order to resolve hangs because unpack was lagging behind pack and math TRISCs, `test_tilize_calculate_utilize.py` started to fail because it attempted to fuse two LLK pipeline runs. This means that in the first run we unpack from two buffers in L1 and after math we pack to two other buffers in L1. Then in the second run we unpack from these two buffers that pack wrote to, however `unpacker` started to read from the buffers before the `packer` completed writing to L1. 

This worked previously because the `unpacker` was severely lagging behind `packer` and `math` (approximately 700 cycles), and this lag allowed for `packer` to finish writing to L1 before the `unpacker` would start reading. Since `unpacker` is synced with `packer` we must make `unpacker` wait for packer to finish before it begins reading from L1.

### What's changed
Implemented semaphore/ flag to control the execution flow required between different components of the system. In `tilize_calculate_untilize_L1.cpp` file we use `packer` mailbox as a semaphore flag for inter-process communication mechanism with the `unpacker`. The mailbox is a shared memory location that packer processor writes to, and unpacker processor reads from. The packer will update the value it stores in its mailbox to signal that it has written to L1. Meanwhile the unpacker will wait/stall until packer signals that it has finished before it begins.

The use of `__atomic_load_n` with `__ATOMIC_SEQ_CST` means that the value is being read atomically with a memory synchronization order, ensuring data consistency and avoiding race conditions without the need for complex locking mechanisms.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/sources/tilize_calculate_untilize_L1.cpp
+++ b/tests/sources/tilize_calculate_untilize_L1.cpp
@@ -44,8 +44,24 @@ void run_kernel()
     _llk_unpack_tilize_init_(UNPACK_B_IN, UNPACK_B_OUT, 1, FACE_R_DIM, false);
     _llk_unpack_tilize_(L1_ADDRESS(buffer_B), 0, UNPACK_B_IN, 1, FACE_R_DIM, 4, false);
 
-    t6_semaphore_wait_on_zero<p_stall::STALL_SYNC>(semaphore::PACK_DONE);
-    t6_semaphore_get<>(semaphore::PACK_DONE);
+    /*
+    In this test we fuse two LLK pipeline runs, one is to unpack untilized buffers/operands from L1 (39-45) and pack them in tilized format(130-145).
+    The next run unpacks these two tilized operands, performs a math compute and pack them out in utilized format.
+    Since we have set all three TRISCs to run at the same time, fusing these two runs will cause a race condition where unpacker will immediately read from
+    L1 before the packer has completed writing to L1. To prevent the unpacker from prematurely reading from L1 before packer has completed write
+    the unpacker needs to wait for packer to finish writing to L1 before it starts reading from L1 for the second iteration of LLK pipeline.
+
+    Synchronization is accomplished between the packer and unpacker operations using the PACK_DONE semaphore.
+    The packer first writes data to L1 and signals the unpacker by incrementing the semaphore (PACK_DONE = 1).
+    The unpacker waits for the semaphore to be set to 1 before reading the data from L1.
+    This ensures that the unpacker does not read premature or incorrect data, preventing data race conditions.
+    Once the unpacker starts reading, it decrements the semaphore (PACK_DONE = 0) signalling it has started processing data.
+    */
+
+    t6_semaphore_wait_on_zero<p_stall::STALL_SYNC>(
+        semaphore::PACK_DONE); // Unpacker waits on signal when packer will increment semaphore to 1 (waits while semaphore == 0), utilizing SEMWAIT.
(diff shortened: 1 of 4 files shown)


tests/sources/tilize_calculate_untilize_L1.cpp
```diff
@@ -25,6 +25,7 @@ volatile uint32_t* const buffer_B = reinterpret_cast<volatile uint32_t*>(0x1b000
 
 volatile uint32_t* const buffer_A_tilized = reinterpret_cast<volatile uint32_t*>(0x1c000);
 volatile uint32_t* const buffer_B_tilized = reinterpret_cast<volatile uint32_t*>(0x1d000);
+volatile std::uint32_t* const mailbox = reinterpret_cast<volatile std::uint32_t*>(0x19FF4);
```
fvranicTT, line 28: we could use C++ atomics instead:
```cpp
std::atomic<std::uint32_t* const> mailbox = reinterpret_cast<std::uint32_t*>(0x19FF4);
```

while becomes:
```cpp
while (*mailbox.load(std::memory_order_acquire) != 0x6) {
    std::this_thread::yield(); // Yield to reduce CPU usage, not sure if it will work on our cores?
}
```
And store becomes:
```cpp
*mailbox.load(std::memory_order_relaxed) = 0x6;
```

nvelickovicTT: Have you tried the STALLWAIT we discussed? I think this approach introduces to big of an overhead, because we use TRISC to synchronize Tensix threads.
This is fine as a way to verify that conceptually the issue is this synchronization, but in runtime we should avoid this way of synchronization.

skotaracTT: I agree! I updated the solution to use Tensix thread semaphore `t6_semaphore_get` and `t6_semaphore_post` defined in `ckernel.h`. They call `TTI_STALLWAIT` and `TTI_SEMGET`. We could also use `TT_STALLWAIT` but `TTI_STALLWAIT` is less overhead since it resolves as compile-time constants and embeds the corresponding Tensix instruction into the instruction stream.

tests/sources/tilize_calculate_untilize_L1.cpp
```diff
@@ -13,6 +13,7 @@
 uint32_t unp_cfg_context          = 0;
 uint32_t pack_sync_tile_dst_ptr   = 0;
 uint32_t math_sync_tile_dst_index = 0;
+constexpr uint8_t MUTEX_INDEX     = 0;
```
fvranicTT, line 16:
```suggestion
constexpr uint8_t SEMAPHORE_INDEX     = 0;
```

tests/sources/tilize_calculate_untilize_L1.cpp
skotaracTT: > Have you tried the STALLWAIT we discussed? I think this approach introduces too big of an overhead, because we use TRISC to synchronize Tensix threads. This is fine as a way to verify that conceptually the issue is this synchronization, but in runtime we should avoid this way of synchronization.

I ended up incorporating `TTI_SEMWAIT` called in `t6_semaphore_wait_on_zero` in order for unpacker to wait while semaphore is 0 and when semaphore is incremented by packer the function returns and we can begin to unpack. 
However the function name is counterintuitive and I used t6_semaphore_wait_on_max` instead due to confusion (which prevented the synchronization from working). So I opened an issue #71 to rename the functions for better readability and future use.

nvelickovicTT (approved):
Just add some comments on what and why the sync is needed. Otherwise, all good.

PR #74, opened 2025-04-01
build: use docker images

fvranicTT:
### Ticket
None

### Problem description
Our build pipeline hasn't been using Docker for now.

### What's changed
This PR adds docker setup and configuration to run our testing infra in containers.
On top of it, this PR contains definition of IRD Docker file.

### Type of change


tests/setup_sfpi.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
nvelickovicTT, line 1: The only part I didn't understand is why can't we ingest sfpi in the same way as tt-smi or exalens?

nvelickovicTT (approved):

tests/setup_sfpi.sh
fvranicTT: We can and we did. I can find the commit where I changed it. It was one of the last ones. The reason is frequent change of sfpi version which would result in building docker image every week. I was trying to reduce the frequency of that.

fvranicTT: See the last commit: https://github.com/tenstorrent/tt-llk/pull/74/commits/bb3e8e837a17f8784719f63b7f71e82e5e560314

fvranicTT: The idea is to fetch sfpi via CPM when we add CMake support. Just like any other dependency

PR #82, opened 2025-04-02
[Test Infra] Ensure Tensix Operations Complete Before Reading Results from L1 & Wait for Tensix Completion to Expose Hangs

skotaracTT:
### Problem description
Previously, we were reading from L1 before verifying if our Tensix operations completed. This could have resulted in reading invalid data from L1, leading to misinterpretation of test failures as functional issues, when in fact the problem might have been a synchronization issue or a hang.
### What's changed
To address this, we’ve updated the flow to ensure that TRISC calls are fully completed before we read the results from L1. Additionally, we replaced our assertion calls with wait calls that pause execution until the Tensix operation finishes. This change improves our ability to distinguish between actual test failures (due to incorrect results) and potential hangs, as the test will now wait for the operation to complete instead of failing immediately on an assertion, which previously didn’t provide visibility into how long the assertion had been false.

### Type of change
- [ ] Documentation update

### DIFF
--- a/tests/python_tests/helpers/device.py
+++ b/tests/python_tests/helpers/device.py
@@ -131,23 +131,31 @@ def get_result_from_device(
         raise ValueError(f"Unsupported format: {formats.pack_dst}")
 
 
-def assert_value_with_timeout(core_loc, mailbox_addr, timeout=0, poll_interval=0.1):
-    end_time = time.time() + timeout
+def wait_until_tensix_complete(core_loc, mailbox_addr, timeout=30, max_backoff=5):
+    """
+    Polls a value from the device with an exponential backoff timer and fails if it doesn't read 1 within the timeout.
 
-    while time.time() < end_time:
-        if read_word_from_device(core_loc, mailbox_addr) == 1:
-            return True
-        time.sleep(poll_interval)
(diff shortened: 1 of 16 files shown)


nvelickovicTT (approved):

tests/python_tests/helpers/device.py
```diff
@@ -131,23 +131,20 @@ def get_result_from_device(
...
-    while time.time() < end_time:
-        if read_word_from_device(core_loc, mailbox_addr) == 1:
-            return True
-        time.sleep(poll_interval)
+    start_time = time.time()
+    print(f"Waiting for {mailbox_addr.name}\n")
+    while read_word_from_device(core_loc, mailbox_addr.value) != 1:
+        if time.time() - start_time > 30:
+            print(f"Timeout reached: waited 30 seconds for {mailbox_addr.name}")
+            break
```
fvranicTT, line 141: Shouldn't we assert(False) here and fail?

skotaracTT: Yup! Fixed it to raise an assertion after 30 seconds of waiting.

fvranicTT (approved):

PR #85, opened 2025-04-03
LLK tilize support for Int32/UInt32 inputs

pmilenkovicTT:
### Ticket
[#16860](https://github.com/tenstorrent/tt-metal/issues/16860)
### Problem description
Int32/UInt32 tilize support was missing on both WH and BH.

### What's changed
In order to tilize 32-bit inputs without losing accuracy, the data cannot be unpacked to Source A/Source B registers - the provided solution unpacks tilized 32-bit integers directly into the destination register. 

### Type of change

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
@@ -27,14 +27,14 @@ inline void _llk_unpack_tilize_mop_config_(const bool narrow_tile = false, const
     static constexpr uint unpack_srca =
         TT_OP_UNPACR(SrcA, 0b1 /*Z inc*/, 0, 0, 0, 1 /* Set OvrdThreadId*/, 1 /*Set Dvalid*/, p_unpacr::RAREFYB_DISABLE, 0, 0, 0, 0, 1);
     static constexpr uint unpack_srca_to_dest =
-        TT_OP_UNPACR(0, 0b00010001 /*Z inc*/, 0, 0, 0, 1 /* Set OvrdThreadId*/, 1 /*Set Dvalid*/, p_unpacr::RAREFYB_DISABLE, 0, 0, 0, 0, 1);
+        TT_OP_UNPACR(0, 0b00010001 /*Z inc*/, 0, 0, 0, 1 /* Set OvrdThreadId*/, 0, p_unpacr::RAREFYB_DISABLE, 0, 0, 0, 0, 1);
     static constexpr uint unpack_srcb_set_dvalid = TT_OP_UNPACR_NOP(SrcB, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
 #endif
     if (is_32bit_integer)
     {
         const uint32_t outerloop     = 1;
         constexpr uint32_t innerloop = 1;
-        ckernel_template tmp(outerloop, innerloop, unpack_srca);
+        ckernel_template tmp(outerloop, innerloop, unpack_srca_to_dest);
         tmp.program(instrn_buffer);
     }
     else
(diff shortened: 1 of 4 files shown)


tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
```diff
@@ -16,23 +16,35 @@
...
+        TT_OP_UNPACR(0, 0b00010001 /*Z inc*/, 0, 0, 0, 1 /* Set OvrdThreadId*/, 0, p_unpacr::RAREFYB_DISABLE, 0, 0, 0, 0, 1);
     static constexpr uint unpack_srcb_set_dvalid = TT_OP_UNPACR_NOP(SrcB, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
 #endif
-
-    const uint32_t outerloop     = 1;
-    constexpr uint32_t innerloop = 1;
-    ckernel_template tmp(outerloop, innerloop, unpack_srcb_set_dvalid);
-    tmp.set_start_op(unpack_srca);
-    tmp.program(instrn_buffer);
+    if (is_32bit_integer)
```
fvranicTT, line 33: This is a bit condensed version:
```cpp
const uint32_t outerloop = 1;
const uint32_t innerloop = 1;

ckernel_template tmp(
    outerloop,
    innerloop,
    is_32bit_integer ? unpack_srca_to_dest : unpack_srcb_set_dvalid
);

if (!is_32bit_integer)
{
    tmp.set_start_op(unpack_srca);
}

tmp.program(instrn_buffer);
```
Note: I made innerloop const (vs constexpr), to make it aligned with the outerloop, as in this case it doesn't matter if it is constexpr or const.

tt_llk_wormhole_b0/common/inc/cmath_common.h
```diff
@@ -232,6 +232,10 @@ inline constexpr bool is_32bit_input(const std::uint32_t src_format, const std::
 {
     const uint input_df  = src_format & 0xF;
     const uint output_df = dst_format & 0xF;
+    if (src_format == (uint32_t)DataFormat::UInt32)
```
fvranicTT, line 235: We need to stop adding C-style casts to the codebase, it's the unsafest way of casting in C++. Not your fault, it's everywhere. Instead:
```cpp
if (src_format == static_cast<std::underlying_type_t<DataFormat>>(format))
```
Looks scary, but if we ever change DataFormat's underlying type, we don't have to revisit this change.

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
```diff
@@ -57,7 +69,8 @@ inline void _llk_unpack_tilize_init_(
     const std::uint32_t unpack_dst_format = 0,
     const std::uint32_t ct_dim            = 0,
     const std::uint32_t face_r_dim        = FACE_R_DIM,
-    const bool narrow_tile                = false)
+    const bool narrow_tile                = false,
+    const bool is_32bit_integer           = false)
```
nvelickovicTT, line 73: You shouldn't need this parameter. You have src format already and can use `is_32bit_input()` function.

tt_llk_wormhole_b0/common/inc/cmath_common.h
```diff
@@ -232,6 +232,10 @@ inline constexpr bool is_32bit_input(const std::uint32_t src_format, const std::
 {
```
nvelickovicTT, line 232: This function has a misleading name. Mentions only input even though it check the format of both input and output.
Can you add a todo with a link to an issue to correct this in a comment above the function, please.

tt_llk_wormhole_b0/common/inc/cmath_common.h
nvelickovicTT, line 235: Is it enough to check the src format? Or we need to do it for dst as well? This function needs some cleanup, I guess.

tt_llk_wormhole_b0/llk_lib/llk_unpack_tilize.h
```diff
@@ -106,40 +107,93 @@ inline void _llk_unpack_tilize_(
     // Program srcA and srcB base addresses
     std::uint32_t num_loops = narrow_tile ? 2 : num_faces / 2;
 
-    for (std::uint32_t n = 0; n < num_loops; n++)
+    if (!is_32bit_integer)
```
nvelickovicTT, line 110: This function got a little messy here. Let's separate the functionality in this if/else into 2 functions. Something like `*unpack_tilize() `and `*unpack_to_dest_tilize()` and call these depending on which situation we have.
And I think you should be able to instrument the code to not need this flag, and use `is_32bit_input()`.

tt_llk_wormhole_b0/llk_lib/llk_unpack_tilize.h
```diff
@@ -106,40 +107,93 @@ inline void _llk_unpack_tilize_(
...
+            TTI_STALLWAIT(p_stall::STALL_TDMA, p_stall::UNPACK);
+
+            // Don't clear the CH1 W counter - needed for multiple tiles
+            TTI_SETADCZW(0b001, 0, 0, 0, 0, 0b1011);
+
+            // Increment address to point to bottom faces in L1
+            address += bot_face_offset_address;
+
+            // Get tile address
+            cfg[THCON_SEC0_REG3_Base_address_ADDR32] = address;
```
nvelickovicTT, line 184: Is the verdict that we don't want to use multiple contexts here? Or is it just a leftover from debugging?

nvelickovicTT (changes requested):

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
pmilenkovicTT: Done, removed the flag

tt_llk_wormhole_b0/common/inc/cmath_common.h
pmilenkovicTT: We still don't have the functionality to convert 32-bit integers into 16-bit like we have for Float32 type - that's why I think that it is enough to check only the source (dest format will always be the same as the source in case of Int32/Uint32).

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
pmilenkovicTT: Added it, thank you!

tt_llk_wormhole_b0/common/inc/cmath_common.h
pmilenkovicTT: Removed the whole line as it ended up being unnecessary, but applied static_cast in the other parts of the code. Thanks for the tip!

tt_llk_wormhole_b0/llk_lib/llk_unpack_tilize.h
pmilenkovicTT: Will add it as a separate issue to implement it with multiple contexts.

nvelickovicTT (approved):

fvranicTT (approved):

tt_llk_wormhole_b0/common/inc/cmath_common.h
pmilenkovicTT: Created [a new issue for this](https://github.com/tenstorrent/tt-llk/issues/95).

PR #86, opened 2025-04-03
fix: remove superfluous make clean calls

fvranicTT:
### Ticket
None

### Problem description
There are superfluous make clean calls after execution of every test. These aren't needed and present a leftover from debugging.

### What's changed
This PR removes them.

### Type of change


skotaracTT (commented):
If we no longer call `run_shell_command` we could remove it from `__init__.py` for helper imports until we find the need to call it again. For now we can keep the function defined. The only place we now call run_shell_command is in `hardware_controller.py` which imports directly from utils.

skotaracTT (approved):
If tests pass as expected on WH and BH arch then this is a great cleanup. Reduces slow down and removes unnecessary calls👌

fvranicTT:
> If we no longer call `run_shell_command` we could remove it from `__init__.py` for helper imports until we find the need to call it again. For now we can keep the function defined. The only place we now call run_shell_command is in `hardware_controller.py` which imports directly from utils.

Good thinking. We don’t use hw controller anymore, so I will remove it, too.

PR #97, opened 2025-04-08
Remove race conditions in packer thread.

amahmudTT:
### Ticket

https://github.com/tenstorrent/tt-metal/issues/18779#issuecomment-2782834310

### Problem description
Matmul hangs non deterministically on P100 scrappy boards but not production boards. This PR fixes the issue of packer thread updating the register containing an address used by the packer before the packer mop finishes. The change added a wait to the packer in an existing STALLWAIT and a NOP after a WRCFG (without which another race condition may be there). I tested the code without the added nop to make sure timing was the same as the code without this PR, and I found the issue to be resolved wheras the original code hanged. This proves the presence of the race condition, and moreover when the test passed in debug mode I saw PCC values near 0 which also shows garbage values got written as output by the packer. 

Some explanation as to why this race happens and we see it in Blackhole from the original issue :

matmul_pack_tile calls llk_matmul_pack to pack multiple tiles together. llk_matmul_pack looks like this

```
template <bool out_of_order_output = false, bool untilize = false, bool is_fp32_dest_acc_en = false>
inline void llk_matmul_pack(
   ---- REMOVED ---- 
    for (uint32_t tile_index = start_tile_index; tile_index < start_tile_index + ntiles; tile_index++) {
        std::uint32_t pack_tile_addr =
            get_output_tile_address<out_of_order_output, untilize>(output_id, output_tile_index);

        _llk_pack_<DST_SYNC_MODE, untilize, is_fp32_dest_acc_en>(tile_index, pack_tile_addr);
    }
}
```

llk_pack looks like this

```
    program_packer_destination(address);

    mop_run(1, 1);

    TT_SETADCZW(p_setadc::PAC, 0, 0, 0, 0, 0b0101); // reset z counters
```

Now the mop is going to do the packing and given the whole thing is in a for loop, it is not guaranteed that when we go back llk_pack again the previous pack is finished or not as all the pack data are synchronized in cb_push_back which is called after the top level function is completed. So in theory we could program the destination address in program_packer_destination before the previous packing is finished as we are wrtiting to a register using WRCFG and there is no WRCFG type of instruction in the MOP to make sure this one has a chance of stay behind the pack instructions (even then it might sneak in). When L1 accumulation is ON packer will also read data from L1 and combine with its own data and write to L1, so what data it reads also will decide whether it will hang or not when it writes to the wrong address (my explanation for seeming data dependency for hangs). In large grid matmuls packer_L1_accumulation is enabled which might slow down packing since the L1 will be read more, and since the UNPACKER also reads, things may slow down causing this race. 

Now why would this not manifest in Wormhole ? In Wormhole we update the register using REG2FLOP which is a THCON instruction (although there is a question asked as comments in the assembly asking whether it is THCON or CFG). If it is a THCON instruction, the MOP has a THCON instruction as the ending OP, so if that instruction is issued by the MOP before this one then until and unless that is finished the REG2FLOP cannot execute due to ordering for the same EXU. On top of that in Wormhole the bandwidth of the Unpacker is much lower, thus it it isl ikely it will compete less with the Packer, in case the bandwidth from L1 matters when accumulation to L1 is done by Packer.

**_But in the end of the day I think it is theoretically possible to artificially contrive a situation where this hang is manifested in Wormhole, but then again should we wiat for it to manifest ?_**

### What's changed
Fixed program_packer_destination by making it wait for the Packer to finish before updating address. 
Also fixed program packer_dest_offset_register function even though it is is not used.

### Type of change

- [ ] Documentation update

### Checklist

- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml) CI passes (if applicable)

### DIFF
--- a/tt_llk_blackhole/common/inc/cpack_common.h
+++ b/tt_llk_blackhole/common/inc/cpack_common.h
@@ -464,14 +464,21 @@ inline void select_packer_dest_registers()
 // Program packer destination addresses from GPRs
 inline void program_packer_destination(uint32_t addr)
 {
+    /*
+       The GPR OUTPUT_ADDR is only used by the packer mop when writing tile headers.
+       Since we do not write tile headers in tt-metal, we do not need to wait for
+       packer to finish or say put a stallwait at this point.
+       We just need to make sure we wait before issuing the WRCFG.
+    */
     uint32_t new_l1_addr = (1 << 31) | addr;
     TT_SETDMAREG(0, LOWER_HALFWORD(addr), 0, LO_16(p_gpr_pack::OUTPUT_ADDR));
     TT_SETDMAREG(0, UPPER_HALFWORD(new_l1_addr), 0, HI_16(p_gpr_pack::OUTPUT_ADDR));
 
-    TTI_STALLWAIT(p_stall::STALL_CFG, p_stall::THCON);
+    TTI_STALLWAIT(p_stall::STALL_CFG, p_stall::THCON | p_stall::PACK);
     TTI_WRCFG(p_gpr_pack::OUTPUT_ADDR, 0, THCON_SEC0_REG1_L1_Dest_addr_ADDR32);
 
     TT_SETDMAREG(0, UPPER_HALFWORD(addr), 0, HI_16(p_gpr_pack::OUTPUT_ADDR));
+    TTI_DMANOP; // One NOP should be enough for WRCFG due to SETDMAREG above.
 }
 
 // RT: If multiple contexts are used, for issue #https://github.com/tenstorrent/tt-llk-bh/issues/20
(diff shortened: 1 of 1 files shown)


ncvetkovicTT (approved):
lgtm

amahmudTT:
All post commit tests pass.
https://github.com/tenstorrent/tt-metal/actions/runs/14343605229
https://github.com/tenstorrent/tt-metal/actions/runs/14341130894

ttmtrajkovic (approved):

amahmudTT:
> @amahmudTT, I read your description and I have 2 comments:
> Milos

Talked with Marco, as for the latching, he said yes it supposed to get latched properly. But there are many things that can go wrong here and he needs to look into it very deeply here. For example if the mop does not let WRCFG to finish while it is running, will the next MOP be able to sample it in time ? The STALLWAIT previously in place for WRCFG was waiting on a non-ideal EXU to mitigate the proper wait it needed (before my code change). All these issues need a much deeper look but he is currently not available for that. But in the mean time given my code change is working and there is an issue for the Tensix team to look into it we should be good. I need to make a test case that hits this issue so that it is prevented in future code and architecture.

ttmtrajkovic:
thanks @amahmudTT.

I am ok with everything you wrote, with caveat that Marco and the hw team should work on generating a test case for this. That isn't software requirement.


milos

amahmudTT:
> thanks @amahmudTT.
> 
> I am ok with everything you wrote, with caveat that Marco and the hw team should work on generating a test case for this. That isn't software requirement.
> 
> milos

FYI , the ask for the test case was from Paul though, not from HW team, I have not discussed with them regarding test case. I will let Paul know.

PR #101, opened 2025-04-09
feat: add support for more sfpu unary ops

fvranicTT:
### Type of change

- [ ] Documentation update

### Checklist


tests/python_tests/test_eltwise_unary_sfpu.py
```diff
@@ -34,7 +38,15 @@ def generate_golden(operation, operand1, data_format):
     all_format_combos,
     dest_acc=[DestAccumulation.No, DestAccumulation.Yes],
     approx_mode=[ApproximationMode.No, ApproximationMode.Yes],
-    mathop=[MathOperation.Sqrt, MathOperation.Log, MathOperation.Square],
+    mathop=[
```
skotaracTT, line 41: Same can be done for mathop map 😁
```suggestion
        @classmethod
        def operations(cls):
            return [op for op in cls if op._function is not None]
```
Here's an example
`mathop = MathOperation.operations()`

fvranicTT: In principle, I agree, but at the moment our MathOperation contains a mix of unary and binary ops. And these can't be mixed. Maybe we should split the ops first?

skotaracTT: yes I agree as well, sounds good!

nvelickovicTT (approved):
Please just keep the old implementation for some time to be able to review it and do experiments.
Add comment to describe what was wrong with it and why it is till there.

PR #110, opened 2025-04-15
[Test Infrastructure] Fixing how data is read from L1 and written to L1 (in bytes)

skotaracTT:
### Problem description
There is a reoccurring issue where data written to L1 is not organized in same order as when packer writes data to L1, in other words our program was writing incorrectly to L1 and this caused comparison of golden and result to be incorrect even though the numbers produced were accurate they were out of order. 

Once this issue was addressed and solved, our data was in stored uniform order throughout, however we were reading words from L1 using tt-exalens but misinterpreting in which order to read the bytes, and according to which format. Since this would cause too much logic and work around to develop our own model for how we read bytes from a word in L1, instead we now use tt-exalens function `read_from_device` which reads bytes from words in L1 correctly.

Now there is no need for bug patches made prior when reading from device in certain formats like bfp8_b or float16, everything is uniform and bytes are read in consecutive order for each format except the number of bytes varies according to size of format/datum. 

### What's changed
In `device.py` instead of calling `read_words_from_device` and from there extracting bytes and interpreting them further, instead we now call function `read_from_device` from exalens which gives us a string of bytes that we later convert to a list of bytes in `unpack.py` and convert to wanted (pack_dst) format accordingly. 

### Type of change

- [ ] Documentation update


tests/python_tests/helpers/device.py
```diff
@@ -22,10 +22,8 @@ def collect_results(
     core_loc: str = "0,0",
     sfpu: bool = False,
 ):
-    read_words_cnt = calculate_read_words_count(formats.pack_dst, sfpu)
```
lpremovicTT, line 25: Can we also remove these functions `calculate_read_words_count` and `flatten_list` if they are not used anywhere else?

tests/python_tests/helpers/unpack.py
```diff
@@ -8,69 +8,40 @@
...
         bytes_data = bytes(byte_list[:2])
         unpacked_value = struct.unpack(">e", bytes_data)[0]
         return unpacked_value
 
     limited_packed_list = packed_list[:2048]
-    result = [
-        bytes_to_float16(limited_packed_list[i : i + 2])
+
+    return [
+        bytes_to_float16(limited_packed_list[i : i + 2][::-1])
```
fvranicTT, line 20: What if we were to use [reversed()](https://docs.python.org/3/library/functions.html#reversed) in line 14:
```python
unpacked_value = struct.unpack(">e", reversed(bytes_data))[0]
```
In that case we wouldn't need to flip the bytest two by two and create temporary lists every time. Reversed returns an iterator and should be more performant, but it's also more expressive, which is what matters more (to me) here.

tests/python_tests/helpers/unpack.py
lpremovicTT, line 1: Was int32 unpack checked, or is that also for another pr?

tests/python_tests/helpers/unpack.py
```diff
@@ -20,48 +14,25 @@ def bytes_to_float16(byte_list):
...
-    if unpack_src == DataFormat.Bfp8_b and pack_dst != unpack_src:
-        for i in range(0, len(result), 2):
-            result[i], result[i + 1] = result[i + 1], result[i]
-    return result
-
 
 def unpack_bfp16(packed_list, unpack_src, pack_dst):
     def bytes_to_bfloat16(byte_list):
-        bytes_data = bytes(byte_list[:2] + [0, 0])  # Ensure we include padding
+        bytes_data = bytes(list(byte_list[:2]) + [0, 0])  # Ensure we include padding
```
lpremovicTT, line 26: Can we settle on where is input tensor and a single datum truncated to the correct size. both bytes_to_x and unpack_x perform those truncations, if that is how we want it can they be also added for float32 and int32. Although i would also argue that bytes_to_x should require that len(byte_list) == DATUM_SIZE at input as that is already done in unpack_x so no need to do the same thing twice. As for input tensor truncation (line 30) it should be done outside the unpack_x as unpack_x should work for any number of datums in a tensor.

tests/python_tests/helpers/pack.py
lpremovicTT, line 1: I assume most formats are now packed properly and there don't seem to be superfluous inversions. But there are still numerous calls to `int_to_bytes_list`, are we looking to cleanup pack in a future PR?
Also pack refers to float as fp32 while unpack refers to it as float32, just note to refactor in a future PR if not now.

tests/python_tests/helpers/unpack.py
skotaracTT: seems to work!

tests/python_tests/helpers/device.py
skotaracTT: Yes of course. This is also accomplished in the next PR. Thanks for the reminder 😊

tests/python_tests/helpers/unpack.py
skotaracTT: It wasn't in the PR

tests/python_tests/helpers/unpack.py
skotaracTT: Yes I completely agree! This would definitely cleanup repetition and make code more readable and easier to follow. The input tensor truncation is accomplished in the next PR which aims to dynamically define the size of the input tensor depending on the size of the datum.

tests/python_tests/helpers/pack.py
skotaracTT: Yes this function will not longer be needed once we start dynamically defining the size of our tensor according to datum size (next PR). Thus it will be cleanup and removed from the infrastructure.
Nice catch for the fp32, I'll fix it for this PR 🤝

PR #111, opened 2025-04-15
[Test Infra] Added flexibility for dynamic byte count calculation for reading datums of different sizes

skotaracTT:
### Problem description

Instead of reading words from device to extract our result vector, we instead call read_from_device which reads the data in bytes. Previously this was set to read a fixed number or words and then a fixed number of bytes (maximum amount of bytes because it was set to read bytes for 32bit datums). 
However, we want to be able to be flexible with our stimuli generation for testing, which means input tensors will be of different sizes and different formats, thus the number of bytes we read should be flexible to these changes, instead of fixed. 

### What's changed

Now we read the exact number of bytes necessary, we take into account the size of our output vector and the size of our datum, defined by `format_num_bytes` map. This calculation is done in `utils.py::calculate_read_byte_count` and the result is passed into exalens' `read_from_device` function called in `device.py`.
### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/helpers/device.py
+++ b/tests/python_tests/helpers/device.py
@@ -15,6 +15,8 @@ import inspect
 import time
 from helpers.param_config import *
 
+MAX_READ_BYTE_SIZE_16BIT = 2048
+
 
 def collect_results(
     formats: FormatConfig,
@@ -105,7 +107,6 @@ def get_result_from_device(
     read_data_bytes: bytes,
     core_loc: str = "0,0",
     sfpu: bool = False,
-    num_tiles: int = 1,
 ):
     # Dictionary of format to unpacking function mappings
     unpackers = {
@@ -120,14 +121,6 @@ def get_result_from_device(
         unpack_func = unpack_bfp16 if sfpu else unpack_bfp8_b
     else:
         unpack_func = unpackers.get(formats.pack_dst)
-
-    if formats.pack_dst in [
-        DataFormat.Float16_b,
-        DataFormat.Float16,
-    ]:  # and num_tiles > 1:
-        read_data_bytes = read_data_bytes[
-            :2048
-        ]  # truncating tensor size so that we only read one tile at a time
 
     if unpack_func:
         num_args = len(inspect.signature(unpack_func).parameters)
--- a/tests/python_tests/helpers/device.py
+++ b/tests/python_tests/helpers/device.py
@@ -18,11 +18,14 @@ from helpers.param_config import *
 
 def collect_results(
     formats: FormatConfig,
+    tensor_size: int,
     address: int = 0x1C000,
     core_loc: str = "0,0",
     sfpu: bool = False,
 ):
-    read_data = read_from_device(core_loc, address, num_bytes=4096)
+
+    read_bytes_cnt = calculate_read_byte_count(formats, tensor_size, sfpu)
+    read_data = read_from_device(core_loc, address, num_bytes=read_bytes_cnt)
     res_from_L1 = get_result_from_device(formats, read_data, sfpu)
     return res_from_L1
 
@@ -102,6 +105,7 @@ def get_result_from_device(
     read_data_bytes: bytes,
     core_loc: str = "0,0",
     sfpu: bool = False,
+    num_tiles: int = 1,
 ):
     # Dictionary of format to unpacking function mappings
     unpackers = {
@@ -116,6 +120,14 @@ def get_result_from_device(
         unpack_func = unpack_bfp16 if sfpu else unpack_bfp8_b
     else:
         unpack_func = unpackers.get(formats.pack_dst)
+
+    if formats.pack_dst in [
+        DataFormat.Float16_b,
+        DataFormat.Float16,
+    ]:  # and num_tiles > 1:
+        read_data_bytes = read_data_bytes[
+            :2048
+        ]  # truncating tensor size so that we only read one tile at a time
 
     if unpack_func:
         num_args = len(inspect.signature(unpack_func).parameters)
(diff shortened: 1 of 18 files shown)


tests/python_tests/helpers/device.py
```diff
@@ -117,6 +121,14 @@ def get_result_from_device(
     else:
         unpack_func = unpackers.get(formats.pack_dst)
 
+    if formats.pack_dst in [
+        DataFormat.Float16_b,
+        DataFormat.Float16,
+    ]:  # and num_tiles > 1:
+        read_data_bytes = read_data_bytes[
+            :2048
```
fvranicTT, line 129: Can we put 2048 into a constant and name it? That way it becomes a named constant and not a 🪄 number 😉

tests/python_tests/helpers/utils.py
```diff
@@ -47,14 +47,21 @@ def run_shell_command(command: str):
...
 
-def calculate_read_words_count(format, sfpu=False):
-    if format not in format_sizes:
-        raise ValueError(f"Unsupported format: {format}")
-
-    if sfpu:  # for now just for 16 bit formats
-        return 256 if format in [DataFormat.Float32, DataFormat.Int32] else 128
-
-    return format_sizes[format]
+def calculate_read_byte_count(format: FormatConfig, array_size: int, sfpu=False) -> int:
```
fvranicTT, line 50:
```python
def calculate_read_byte_count(format: FormatConfig, array_size: int, sfpu=False) -> int:
    # Calculate the number of elements that fit in one word (4 bytes per word)
    elements_per_word = 4 // format_num_bytes[format.pack_dst]
    
    # Calculate the total number of words required for the array
    total_words = array_size // elements_per_word
    
    # Calculate the total number of bytes required
    total_bytes = total_words * 4  # Each word is 4 bytes
    
    # Add extra bytes for Bfp8_b format (1 extra byte per block of 4 words)
    if format.pack_dst == DataFormat.Bfp8_b:
        extra_bytes = total_words // 4  # 1 extra byte per block of 4 words
        total_bytes += extra_bytes

    return total_bytes
```

tests/python_tests/helpers/device.py
skotaracTT: of course! If only we could add emojis too 😁👌

fvranicTT: We could, python supports UTF-8 characters 🀄

tests/python_tests/helpers/utils.py
skotaracTT: confirming: you just added spacing (no logic change)?

fvranicTT: Yes, no prints and some spaces (I think I removed 1 intermediate variable).

tests/python_tests/helpers/format_arg_mapping.py
```diff
@@ -68,12 +68,12 @@
     DataFormat.Int32: "PACK_DST_INT32",
 }
 
-format_sizes = {
-    DataFormat.Float32: 1024,
-    DataFormat.Float16: 512,
-    DataFormat.Float16_b: 512,
-    DataFormat.Bfp8_b: 272,
-    DataFormat.Int32: 1024,
+format_num_bytes = {
```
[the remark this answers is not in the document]
skotaracTT: Great!

tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -84,7 +84,7 @@ def test_unary_datacopy(testname, formats, dest_acc):
 
     wait_for_tensix_operations_finished()
     res_from_L1 = collect_results(
-        formats
+        formats, len(src_A)
```
fvranicTT, line 87: might be a nice thing to have formats = formats, tensor_size = len(src_A) in these files. 
Makes it easier to read.

tests/python_tests/helpers/utils.py
skotaracTT: Yeah that's way better!

tests/python_tests/test_tilize_calculate_untilize_L1.py
```diff
@@ -98,7 +98,7 @@ def test_tilize_calculate_untilize_L1(
     wait_for_tensix_operations_finished()
 
     res_from_L1 = collect_results(
-        formats, 0x1E000
```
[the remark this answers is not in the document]
skotaracTT: Noted🫡

tests/python_tests/test_eltwise_unary_sfpu.py
```diff
@@ -78,7 +78,7 @@ def test_eltwise_unary_sfpu(testname, formats, dest_acc, approx_mode, mathop):
 
     wait_for_tensix_operations_finished()
     res_from_L1 = collect_results(
-        formats, sfpu=True
```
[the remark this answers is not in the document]
skotaracTT: Yup I deleted it and all the tests still pass. Good eye!

tests/python_tests/helpers/device.py
```diff
@@ -15,6 +15,8 @@
 import time
 from helpers.param_config import *
 
+MAX_READ_BYTE_SIZE_16BIT = 2048
```
lpremovicTT, line 18: nit: unused

lpremovicTT (approved):

tests/python_tests/helpers/utils.py
```diff
@@ -47,14 +46,11 @@ def run_shell_command(command: str):
...
```
fvranicTT, line 49: We don’t need sfpu param anymore, right?

fvranicTT (approved):

PR #116, opened 2025-04-16
Fix PCK_DEST_RD_CTRL config based on 32 vs 16 bit Dest for int8

rdjogoTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/18914

### Problem description
UInt8 packing is not correctly supported when DEST reg is in 16 bit operating mode, only 32 bit mode. This is because in BBE there was an assumption to always have 32 bit mode enabled for int formats.

### What's changed
Change the programming of PCK_DEST_RD_CTRL config regs based on 32 vs 16 bit Dest for (u)int8.

### Type of change

### Checklist

- [x] [All post commit](https://github.com/tenstorrent/tt-metal/actions/runs/14521751279) CI passes
- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/runs/14521752601) CI passes (if applicable)

### DIFF
--- a/tt_llk_blackhole/common/inc/cpack_common.h
+++ b/tt_llk_blackhole/common/inc/cpack_common.h
@@ -261,10 +261,10 @@ inline void set_packer_config(const uint pack_src_format, const uint pack_dst_fo
     // cfg[THCON_SEC0_REG1_Row_start_section_size_ADDR32+3]=config.val[3];
 
     dest_rd_ctrl_u dest_rd_ctrl;
-    dest_rd_ctrl.val                              = 0;
-    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = (pack_output_src_format == (uint)DataFormat::Int32) | (pack_output_src_format == (uint)DataFormat::Float32) |
-                                                    (is_fp32_dest_acc_en ? 1 : 0);
-    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_int8     = !is_fp32_dest_acc_en && (pack_output_src_format == (uint)DataFormat::Int8);
+    dest_rd_ctrl.val = 0;
+    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data =
+        (pack_output_src_format == (uint)DataFormat::Int32) | (pack_output_src_format == (uint)DataFormat::Float32) | (is_fp32_dest_acc_en ? 1 : 0);
+    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_int8 = !is_fp32_dest_acc_en && (pack_output_src_format == (uint)DataFormat::Int8);
     if (pack_dst_format == (uint)DataFormat::UInt8)
     {
         dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_unsigned = 1;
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/common/inc/cpack_common.h
```diff
@@ -303,9 +304,10 @@ inline void reconfig_packer_data_format(
     TTI_WRCFG(p_gpr_pack::TMP_LO, p_cfg::WRCFG_32b, THCON_SEC0_REG1_Row_start_section_size_ADDR32 + 2);
 
     dest_rd_ctrl_u dest_rd_ctrl;
-    dest_rd_ctrl.val                              = 0;
-    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = (pack_output_src_format == (uint)DataFormat::Int8) | (pack_src_format == (uint)DataFormat::Int32) |
-                                                    (pack_src_format == (uint)DataFormat::Float32) | (is_fp32_dest_acc_en ? 1 : 0);
+    dest_rd_ctrl.val = 0;
+    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data =
+        (pack_output_src_format == (uint)DataFormat::Int32) | (pack_output_src_format == (uint)DataFormat::Float32) | (is_fp32_dest_acc_en ? 1 : 0);
```
nvelickovicTT, line 309: Should this condition cover UInt32, as in WH?

nvelickovicTT (approved):

tt_llk_blackhole/common/inc/cpack_common.h
```diff
@@ -261,9 +261,10 @@ inline void set_packer_config(const uint pack_src_format, const uint pack_dst_fo
     // cfg[THCON_SEC0_REG1_Row_start_section_size_ADDR32+3]=config.val[3];
 
     dest_rd_ctrl_u dest_rd_ctrl;
-    dest_rd_ctrl.val                              = 0;
-    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = (pack_output_src_format == (uint)DataFormat::Int8) | (pack_output_src_format == (uint)DataFormat::Int32) |
-                                                    (pack_output_src_format == (uint)DataFormat::Float32) | (is_fp32_dest_acc_en ? 1 : 0);
+    dest_rd_ctrl.val = 0;
+    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data =
```
fvranicTT, line 265:
```cpp
bool is_32b_format = (pack_output_src_format == (uint)DataFormat::Int32) ||
                     (pack_output_src_format == (uint)DataFormat::Float32);

dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = is_32b_format || is_fp32_dest_acc_en;

dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_int8 = !is_fp32_dest_acc_en && (pack_output_src_format == (uint)DataFormat::Int8);
```

tt_llk_wormhole_b0/common/inc/cpack_common.h
```diff
@@ -248,9 +248,9 @@ inline void set_packer_config(const uint pack_src_format, const uint pack_dst_fo
 
     dest_rd_ctrl_u dest_rd_ctrl;
     dest_rd_ctrl.val                              = 0;
-    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = ((pack_src_format & 0xF) == (uint)DataFormat::Int8) | (pack_src_format == (uint)DataFormat::Int32) |
-                                                    (pack_src_format == (uint)DataFormat::UInt32) | (pack_src_format == (uint)DataFormat::Float32) |
-                                                    (is_fp32_dest_acc_en ? 1 : 0);
+    dest_rd_ctrl.f.PCK_DEST_RD_CTRL_Read_32b_data = (pack_src_format == (uint)DataFormat::Int32) | (pack_src_format == (uint)DataFormat::UInt32) |
```
fvranicTT, line 251: Ditto

tt_llk_blackhole/common/inc/cpack_common.h
rdjogoTT: so `DataFormat::UInt32 == 24` and `DataFormat::Int32 == 8`, which means `DataFormat::UInt32 & 0xF == DataFormat::Int32`

On line 292 we already do `pack_src_format&0xF` to produce `pack_output_src_format`, so this check actually already covers uint32 as well, same as WH

tt_llk_blackhole/common/inc/cpack_common.h
rdjogoTT: would it be a good idea to replace all cases of `(uint)DataFormat*` with `static_cast<DataFormatType>(DataFormat*` then? like everywhere in the code? I don't want to make that change just here if its better

tt_llk_wormhole_b0/common/inc/cpack_common.h
```diff
@@ -362,9 +362,9 @@ inline void reconfig_packer_data_format(const uint pack_src_format, const uint p
...
```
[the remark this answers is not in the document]
rdjogoTT: `&0xF` is extracts the bottom 4 bits of dataformats, and the dataformats were specially selected such that the bottom 4 bits of the uint version of each format is equal to the int version. So is the same as checking for both uint8 and int8

tt_llk_blackhole/common/inc/cpack_common.h
nvelickovicTT: I see. That’s definitely not easy for the reader to figure out. Is there a reason we do it that way?
Since this is not performance critical code, I think it would be better if we check each format explicitly, without applying masks.

tt_llk_blackhole/common/inc/cpack_common.h
fvranicTT: It would. But this change would be enormous.

tt_llk_blackhole/common/inc/cpack_common.h
rdjogoTT: makes sense

fvranicTT (approved):

PR #120, opened 2025-04-22
build: use latest metallium docker image as base for llk ird

fvranicTT:
### Ticket
None

### Problem description
Latest metallium image has been updated. Switching llk ird base to it.

### What's changed
The size of the latest Metallium image has been significantly reduced without affecting functionality (850 MB vs. ~10 GB previously). Starting a new session used to take a few minutes when launching a machine for the first time. The improvement should be noticeable to everyone.

### Type of change


fvranicTT:
Awaiting confirmation from the metal team. With this new image, I can't build tt-metal repo since a lot of dependencies are missing.

Issue #134, opened 2025-04-26
Add support for 32-bit types to _llk_math_transpose_dest_.


jasondavies:
Currently, the transpose_dest implementation uses `TTI_TRNSPSRCB`, and since SrcA/SrcB only support transposing 19 bits, this will fail to work correctly for 32-bit types such as int32 or float32.

jasondavies:
@fvranicTT I'm assuming I need to do this for Blackhole too.  I can copy the implementation over and make a new PR.  Ideally we should add a tt-llk unit test for transpose_dest, which will help check the Blackhole case.

PR #135, opened 2025-04-26
Add support for 32-bit types to _llk_math_transpose_dest_.

jasondavies:
As suggested by @corsix, this performs a transpose on a 16x16 face of 32-bit values by splitting into two faces of 16-bit values (the low 16 bits and high 16 bits), transposing each face separately using `TRNSPSRCB`, and then recombining into 32-bit values.

A neat way to achieve this (also suggested by @corsix!) is to use the `dest_32b_lo` instruction mode flag on `MOVD2B`/`MOVB2D`.  This allows the low bits to be copied to SrcB, transposed, and copied back.  Followed by the same operation on the high bits.

Further optimisation would involve using the full 32 slots in the replay buffer, rather than overwriting slots 16 to 32 as I'm doing currently.

This was just the most straightforward way to get a working proof of concept.

### Ticket

#134

### Problem description

32-bit types aren't supported by transpose_dest.

### What's changed

See above.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
@@ -118,8 +118,8 @@ inline void transpose_dest_configure_mop()
         // Set Sequence[1].
         TTI_SFPCONFIG(0, 5, 0);
 
-        // Reset "misc" macro config with UsesLoadMod0ForStore=1 for all macros.
-        TTI_SFPCONFIG(0x0f0, 8, 1);
+        // Misc: {UsesLoadMod0ForStore=1, WaitForElapsedInstructions=1} for macros 0 and 1.
+        TTI_SFPCONFIG(0x330, 8, 1);
 
         // A 32b face transpose consists of: (movd2b_hi, transpose, movb2d_hi_d2b_lo, transpose, movb2d_lo).
         uint movd2b_hi        = TT_OP_REPLAY(16, 4, 0, 0);
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
@@ -16,8 +16,9 @@
 using namespace ckernel;
 
 // local function declarations
+template <bool is_32bit>
 inline void transpose_dest_configure_addrmod();
-template <int dest_32b_lo>
+template <bool is_32bit>
 inline void transpose_dest_configure_mop();
 
 template <bool is_32bit = false>
@@ -27,28 +28,38 @@ inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 
     TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU);
 
-    ckernel_template::run(instrn_buffer);
-
-    TTI_REPLAY(20, 5, 0, 0);
-    TTI_REPLAY(26, 4, 0, 0);
-
-    math::clear_dst_reg_addr();
-
     if constexpr (is_32bit)
     {
-        transpose_dest_configure_mop<Dest32bLo::On>();
-        ckernel_template::run(instrn_buffer);
-
-        TTI_REPLAY(20, 5, 0, 0);
-        TTI_REPLAY(26, 4, 0, 0);
+#pragma GCC unroll 4
+        for (int i = 0; i < 4; ++i)
+        {
+            TTI_REPLAY(16, 4, 0, 0);
+            TTI_TRNSPSRCB;
+            TTI_REPLAY(20, 4, 0, 0);
+        }
 
         math::clear_dst_reg_addr();
-        transpose_dest_configure_mop<Dest32bLo::Off>();
+
+#pragma GCC unroll 4
+        for (int i = 0; i < 4; ++i)
+        {
+            TTI_REPLAY(24, 4, 0, 0);
+            TTI_TRNSPSRCB;
+            TTI_REPLAY(28, 4, 0, 0);
+        }
+
+        math::clear_dst_reg_addr();
+        ckernel_template::run(instrn_buffer);
+    }
+    else
+    {
+        ckernel_template::run(instrn_buffer);
     }
 
-    TTI_SETRWC(p_setrwc::CLR_AB, 0, 0, 0, 0, p_setrwc::SET_AB);
+    TTI_SETRWC(p_setrwc::CLR_AB, 0, 0, 0, 0, p_setrwc::SET_ABD);
 }
 
+template <bool is_32bit>
 inline void transpose_dest_configure_addrmod()
 {
     addr_mod_t {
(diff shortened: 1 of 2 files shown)


jasondavies:
Incidentally, I can't see any explicit reason why slots 16-32 of the replay buffer are used, instead of 0-16, but I think it's because this allows simultaneous use of the replay buffer by [cumsum](https://github.com/tenstorrent/tt-llk/blob/f9293efa754686e7581b4711168ea5abb05d3594/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_cumsum.h#L151-L167) and transpose_dest.

jasondavies:
I haven't benchmarked, but it's probably going to be faster to issue MOVD2B/MOVB2D instructions directly for the second part of the 32-bit case, rather than rebuild the replay buffer every time and use MOP expansion.  That way the replay buffer can be kept as-is for backwards-compatibility.

fvranicTT:
There was a breaking change in tt-metal workflow (from 3 days ago) that prevents integration tests workflow from running.
I asked tt-metal infra team for help.

jasondavies:
Thanks. BTW, I manually ran `test/tt_metal/unit_tests_llk` and it passed all tests.

jasondavies:
> I haven't benchmarked, but it's probably going to be faster to issue MOVD2B/MOVB2D instructions directly for the second part of the 32-bit case, rather than rebuild the replay buffer every time and use MOP expansion. That way the replay buffer can be kept as-is for backwards-compatibility.

Extracting the MOVD2B/MOVB2D/TRNSPSRCB instructions from the replay/mop config results in 45 instructions.  I'm not sure what the overhead of replay fill + mop config is, but it's going to be at least 16 cycles (replay fill) plus a few cycles for mop config.

There's only really one place this API is used at the moment (the cumsum test in tt-metal), which is probably where the decision to use slots 16-32 of the replay buffer came from.

However, note that `transpose_wh_dest_init_short` is always called before `transpose_wh_dest` in the cumsum test, so it currently always has this additional latency.  This seems unnecessarily wasteful of cycles, unless the primary aim here of replay + mop config is to reduce code size?

jasondavies:
Non-MOP version with `dest_32b_lo=1`:

```
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12);
TTI_TRNSPSRCB;
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 0);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 4);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 8);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_0, p_movb2d::MOV_4_ROWS, 12);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12);
TTI_TRNSPSRCB;
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 0);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 4);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 8);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_0, p_movb2d::MOV_4_ROWS, 12);
TTI_MOVD2B(1, p_movd2b::SRC_ZERO_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0 - 16);
TTI_MOVD2B(1, p_movd2b::SRC_ZERO_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4 - 16);
TTI_MOVD2B(1, p_movd2b::SRC_ZERO_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8 - 16);
TTI_MOVD2B(1, p_movd2b::SRC_ZERO_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12 - 16);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12);
TTI_TRNSPSRCB;
TTI_MOVD2B(1, p_movd2b::SRC_ZERO_OFFSET + 32, ADDR_MOD_2, p_movd2b::MOV_1_ROW, 0); // throwaway to decrement dst
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 0);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 4);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 8);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_0, p_movb2d::MOV_4_ROWS, 12);
TTI_MOVB2D(1, p_movd2b::SRC_ZERO_OFFSET + 0, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 0);
TTI_MOVB2D(1, p_movd2b::SRC_ZERO_OFFSET + 4, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 4);
TTI_MOVB2D(1, p_movd2b::SRC_ZERO_OFFSET + 8, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 8);
TTI_MOVB2D(1, p_movd2b::SRC_ZERO_OFFSET + 12, ADDR_MOD_0, p_movb2d::MOV_4_ROWS, 12);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8);
TTI_MOVD2B(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12);
TTI_TRNSPSRCB;
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 0, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 0);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 4, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 4);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 8, ADDR_MOD_1, p_movb2d::MOV_4_ROWS, 8);
TTI_MOVB2D(1, p_movd2b::SRC_ROW16_OFFSET + 12, ADDR_MOD_0, p_movb2d::MOV_4_ROWS, 12);
```

jasondavies:
This PR can theoretically be merged as-is, since it is backwards-compatible, though it might be good to get feedback on the following options:

1. Modify `transpose_wh_dest_init_short` to fill replay buffer slots 0-31 for the 32b case, and 16-31 when otherwise (for compatibility with the non-32b cumsum test).
2. Remove the use of replay buffers + MOP entirely in favour of unrolled loops that emit the 45-instruction sequence above for the non-32b case, and 90 instructions for the 32b case.

jasondavies:
Unless there's a reason to use MOP/replay here, I think it's better to use unrolled loops to emit the instruction sequence directly, especially as this LLK is probably going to be used in conjunction with other operations that necessitate refilling the replay buffer before every call.

Initialising the replay buffer (16 slots) + MOP costs ~20 cycles of latency, and so the only reason to use it is probably for code size reasons.  Moreover, code size _can_ also be reduced via RISC-V control flow.  I don't know what the latency of these MOVD2A/MOVA2D instructions are, and how much overhead RISC-V control flow would add.

(Not using MOP also makes the code more readable!)

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
```diff
@@ -107,5 +123,5 @@ inline void _llk_math_transpose_dest_init_()
 {
     transpose_dest_configure_addrmod();
 
-    transpose_dest_configure_mop();
+    transpose_dest_configure_mop<0>();
```
nvelickovicTT, line 126: Please use a constexpr for the template param value, to make it more descriptive.
Even better, you can create an enum class, e.g.:

```
enum class DestinationState {
    DEST_32_LO_ON,
    DEST_32LO_OFF
};
```
and then use as

```
transpose_dest_configure_mop<DestinationState::DEST_32_LO_OFF>();
```

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
```diff
@@ -29,9 +32,21 @@ inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
...
+    if constexpr (is_32bit)
+    {
+        transpose_dest_configure_mop<1>();
+        ckernel_template::run(instrn_buffer);
+
+        TTI_REPLAY(20, 5, 0, 0);
+        TTI_REPLAY(26, 4, 0, 0);
+
+        math::clear_dst_reg_addr();
+        transpose_dest_configure_mop<0>(); // TODO use length 32 replay buffer
```
nvelickovicTT, line 46: We will most likely revise this design decision to split replay buffer usage between FPU and SFPU, but for now this is the convention. Replay buffer entries 0-16 are for SFPU, and 17-32 are for FPU.

nvelickovicTT (approved):

nvelickovicTT:
> Unless there's a reason to use MOP/replay here, I think it's better to use unrolled loops to emit the instruction sequence directly, especially as this LLK is probably going to be used in conjunction with other operations that necessitate refilling the replay buffer before every call.
> 
> Initialising the replay buffer (16 slots) + MOP costs ~20 cycles of latency, and so the only reason to use it is probably for code size reasons. Moreover, code size _can_ also be reduced via RISC-V control flow. I don't know what the latency of these MOVD2A/MOVA2D instructions are, and how much overhead RISC-V control flow would add.
> 
> (Not using MOP also makes the code more readable!)

MOP/replay are used for perf reasons. RISCVs used to control Tensix engine would be the bottleneck with issuing Tensix instructions if we didn't use MOP/replay buffers.

You are welcome to do an experiment and compare perf of implementation with and without MOP/replay buffer.

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
jasondavies: I can go the constexpr route.  The enum class makes the code noisier as I would need to introduce integer casts.

jasondavies:
I did some cursory benchmarking and there isn't a massive difference between the two approaches in overall runtime.  Obviously, when using MOP/replay, RISCV does much less work.  I must admit, I was expecting the MOP/replay initialisation to be more visible.

Let's say we stick with MOP/replay, one simple change could be to move `transpose_dest_configure_mop<DEST_32B_LO_OFF>` out of `_llk_math_transpose_dest_init_` and into `_llk_math_transpose_dest_`.  Right now I'm doing this for the 32b case:

- init: configure mop DEST_32B_LO_OFF
- transpose: run MOP, configure mop DEST_32B_LO_ON, run MOP, configure mop DEST_32B_LO_OFF (restores init state)

Instead:
- init: (keep transpose_dest_configure_addrmod)
- transpose: configure mop DEST_32B_LO_OFF, run MOP, configure mop DEST_32B_LO_ON, run MOP

As I mentioned earlier, I think most uses of transpose_dest involve calling init right before the transpose every time anyway, so there would be no difference.  This change simply drops the wasted configure mop DEST_32B_LO_OFF after each call (for 32b case only).

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
jasondavies: Happy to stick with convention for now.

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
```diff
@@ -15,9 +15,15 @@
 
 using namespace ckernel;
 
+constexpr int DEST_32B_LO_OFF = 0;
```
nvelickovicTT, line 18: These could be useful in more places going forward. I suggest to put in https://github.com/tenstorrent/tt-llk/blob/main/tt_llk_wormhole_b0/llk_lib/llk_defs.h and write as:

```
enum class Dest32bLo
{
   Off = 0;
   On = 1;
}
```
and then you can use as `Dest32bLo::Off/Dest32bLo::On`.

nvelickovicTT (approved):

nvelickovicTT:
> I did some cursory benchmarking and there isn't a massive difference between the two approaches in overall runtime. Obviously, when using MOP/replay, RISCV does much less work. I must admit, I was expecting the MOP/replay initialisation to be more visible.
> 
> Let's say we stick with MOP/replay, one simple change could be to move `transpose_dest_configure_mop<DEST_32B_LO_OFF>` out of `_llk_math_transpose_dest_init_` and into `_llk_math_transpose_dest_`. Right now I'm doing this for the 32b case:
> 
> * init: configure mop DEST_32B_LO_OFF
> * transpose: run MOP, configure mop DEST_32B_LO_ON, run MOP, configure mop DEST_32B_LO_OFF (restores init state)
> 
> Instead:
> 
> * init: (keep transpose_dest_configure_addrmod)
> * transpose: configure mop DEST_32B_LO_OFF, run MOP, configure mop DEST_32B_LO_ON, run MOP
> 
> As I mentioned earlier, I think most uses of transpose_dest involve calling init right before the transpose every time anyway, so there would be no difference. This change simply drops the wasted configure mop DEST_32B_LO_OFF after each call (for 32b case only).

I think you are just hitting an edge case of the programming model. The intention is to have the `init` cover the `addr_mod` and `mop` initialization. And than this is called before the actual kernel to set things up and only if there was a different kernel running previously, otherwise no need to configure MOP and addr_mode again.
Here, you are reconfiguring MOP multiple time even inside the kernel, which is something that we are seeing recently with the need for multi phase kernels, e.g. topK and splitting 32bit processing into 2 times 16b.
I would suggest to keep `init` compatible with how the programming model works for most operations, and handle added complexity case by case. If you optimise init for this particular instance, then something else might need to be changed that relies on that init doing MOP config.

jasondavies:
Is there a convention for which `ADDR_MOD_*` can be used by LLKs?  It looks like matmul uses 1-5, for example.

jasondavies:
> Is there a convention for which `ADDR_MOD_*` can be used by LLKs? It looks like matmul uses 1-5, for example.

I'm asking because I'm attempting to rewrite the replay buffer to use fewer instructions, by leveraging ADDR_MOD to auto-increment.

My simple test was as follows:

```diff
+
+    addr_mod_t {
+        .srca = {.incr = 0},
+        .srcb = {.incr = 0},
+        .dest = {.incr = 4},
+    }
+        .set(ADDR_MOD_3);
+
+    addr_mod_t {
+        .srca = {.incr = 0},
+        .srcb = {.incr = 0},
+        .dest = {.incr = -12},
+    }
+        .set(ADDR_MOD_4);
```

And then testing this on just 4 instructions:

```diff
-    TTI_MOVD2B(dest_32b_lo, p_movd2b::SRC_ZERO_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0 - 16);
-    TTI_MOVD2B(dest_32b_lo, p_movd2b::SRC_ZERO_OFFSET + 4, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 4 - 16);
-    TTI_MOVD2B(dest_32b_lo, p_movd2b::SRC_ZERO_OFFSET + 8, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 8 - 16);
-    TTI_MOVD2B(dest_32b_lo, p_movd2b::SRC_ZERO_OFFSET + 12, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 12 - 16);
+    TTI_MOVD2B(dest_32b_lo, 0, ADDR_MOD_3, p_movd2b::MOV_4_ROWS, -16);
+    TTI_MOVD2B(dest_32b_lo, 0, ADDR_MOD_3, p_movd2b::MOV_4_ROWS, -12);
+    TTI_MOVD2B(dest_32b_lo, 0, ADDR_MOD_3, p_movd2b::MOV_4_ROWS,  -8);
+    TTI_MOVD2B(dest_32b_lo, 0, ADDR_MOD_4, p_movd2b::MOV_4_ROWS,  -4);
```

(Note: the src/dst arguments to MOVD2B are a bit confusing.  I assume src means Dest, and dst means SrcB?  I tried the other way round too, just in case.)

My intention is that dest offset should increment by 4 (3 times) and then subtract 12, returning to its original offset so that the reset of the MOP/replay won't be affected.

My question is: why doesn't this work?  What am I missing here?

My plan is of course to do this for SrcB too, but I'm starting small to see if it works.

pcawleyTT:
It might not be the _only_ problem, but specifying a negative value as the last operand of `TTI_MOVD2B` is slightly spicy, as the macro is expecting a value in the range of 0 through 1023 there, which negative numbers aren't. You probably want something like `0x3ff & -16` rather than just `-16` (left as an exercise to the reader: look at the macro definition and figure out what it'll expand to when that operand is plain `-16`).

jasondavies:
Thanks.  I kind of assumed that wouldn't be _too_ much of a problem as the existing code was doing that.  I did however have to use `0x3f & ...` for `.srcb = { .inc = ...}` as the compiler complained!

jasondavies:
I just checked the macro definition and you're right, that is _spicy_!  Pretty sure there may be a deeper issue as I checked `TT_OP_MOVD2B` with/without the `0x3ff` mask and it produced different values.

jasondavies:
I wrote a 32-bit version that uses a 16-slot replay buffer with SrcB for within-face transpose (without MOP), and then an 8-iteration MOP that uses 4 SFPLOAD/SFPSTORE instructions to swap the two 16x16 middle faces.  This is _probably_ better, though it would be nice to have some hard numbers.

jasondavies:
Very helpful, thanks!

jasondavies:
What's the best way to microbenchmark something like this portion of code?

```cpp
ckernel_unpack_template::run(instrn_buffer, 2, 2);
```

Ideally I'd also like the ability to microbenchmark something that presents instructions to both FPU and SFPU, so I'd need some way to synchronise with both FPU and SFPU.  This is to check something like:

1. Replay some number of MOVD2B/TRNSPSRCB/MOVB2D instructions.
2. MOP expansion of some number of SFPLOAD/SFPSTORE instructions.
3. Wait for both FPU and SFPU to complete.

Even something that uses DPRINT is fine to begin with (I can always run different numbers of iterations and check the delta to eliminate any DPRINT overhead).

jasondavies:
Right now I'm using `tensix_sync()` followed by reading `RISCV_DEBUG_REG_WALL_CLOCK_L`, but it would be nice to have a bit more certainty about exactly what I'm measuring.

pcawleyTT:
RE microbenchmarking, there's the RISCV-centric approach:
1. Have RISCV wait for Tensix idle
2. Have RISCV sample the cycle counter
3. Have RISCV fire off the Tensix work you want measured
4. Have RISCV wait for Tensix idle
5. Have RISCV sample the cycle counter again

Or there's the Tensix-centric approach (which still requires a RISCV to fire off the pieces and subsequently collect the results):
1. Have Tensix wait for Tensix idle
2. Have Tensix sample the cycle counter
3. Do the Tensix work you want measured
4. Have Tensix wait for Tensix idle
5. Have Tensix sample the cycle counter

||RISCV-centric|Tensix-centric|
|---|---|---|
|**Wait&nbsp;for&nbsp;Tensix&nbsp;idle**|`tensix_sync()`|`STALLWAIT(STALL_THREAD, what_to_wait_for)`|
|**Sample&nbsp;the&nbsp;cycle&nbsp;counter**<br/>**(Low 32 bits thereof)**|Load from `RISCV_DEBUG_REG_WALL_CLOCK_L` to RISCV GPR|`LOADREG` from `RISCV_DEBUG_REG_WALL_CLOCK_L` to Tensix GPR|

`STALLWAIT` has a range of mode bits, though the result usually boils down to something like "wait until the current thread has completed all previously-issued FPU instructions and SFPU instructions and packer instructions", with the mode bits controlling which set of instructions to wait for completion of. As long as `STALL_THREAD` is specified, no subsequent Tensix instructions from the thread will execute until the conditions are met. `tensix_sync()` is similar, albeit without any mode bits: it waits until the current Tensix thread has completed all previously-issued instructions, and no subsequent RISCV instructions from the thread will execute until then.

There are also options for getting the full 64-bit cycle counter, but as long as the microbenchmark completes in less than 4 seconds, 32 bits should be fine.

jasondavies:
> I wrote a 32-bit version that uses a 16-slot replay buffer with SrcB for within-face transpose (without MOP), and then an 8-iteration MOP that uses 4 SFPLOAD/SFPSTORE instructions to swap the two 16x16 middle faces. This is _probably_ better, though it would be nice to have some hard numbers.

Just pushed my implementation of this (and merged the latest changes from main for the 16-bit version).

I improved the efficiency of the inner 16x16 face swap using `SFPLOADMACRO`, which swaps a pair of 32x32b rows every 2 cycles instead of 4 cycles using the naive approach.  Thanks to @pcawleyTT for invaluable information relating to `SFPLOADMACRO`, without which this would have been virtually impossible!

Compared with my original approach, this only needs to configure the replay buffer and MOP config once, and so eliminates any overhead due to refilling the replay buffer etc.

Additionally, the number of 16x16 movements to and from SrcA/SrcB is now just 2x8, whereas it would have been 2x9 even with the more efficient version that just landed in main.

PR #138, opened 2025-04-28
fix: extend matmul test with Float16 and MathFidelity.[LoFi - HiFi2]

fvranicTT:
### Ticket
None

### Problem description
Extending matmul test with more combinations. Need to check why Float16/DestAccumulation.Yes results in error.

### What's changed
More tests will be running now.

### Type of change

- [ ] Documentation update


skotaracTT:
Good news is that the data format inference model allows for this extension, and checks that the format combination including Float16 works properly by setting intermediate formats and flags in the llk pipeline accordingly for the run to pass. This additional is added in upcoming PR 👍

Issue #80, opened 2025-04-02
Data Format Inference Model


skotaracTT:

skotaracTT:
Data format inference model allows tests to work for more formats and combinations and thanks to this we were able to increase test coverage 4X!

PR #143, opened 2025-04-29
[Test Infrastructure] Data format Inference Model 

skotaracTT:
### Ticket

https://github.com/tenstorrent/tt-llk/issues/80
### Problem description
Tests support Input and Output format parametrization meaning that the user can define the format they what their input tensor in L1 to be and the format of their output in L1. However for the LLK pipeline to run we need to define the intermediary formats including unpacker formats, fpu formats and packer formats. This the user does not define and is left up to us to define according to the best combination which was developed during the format sweep effort. 

### What's changed

The inference model looks at the input and output format defined by client, and dest accumulation, if it is on or off and sets the unpacker and packer src and dest formats accordingly. When dest accumulation is off the packer accepts the format seen in the dst register and then convert to the output formats wanted. However when dest accumulation is on then `Float32` is in dst register and packer gasket can covert to the output format before the data goes into packer for reformatting; thus the packer accepts the source format to be the wanted output format and does no conversion.
- Note that when our wanted output if `Float16_a` and dest accumulation is turned on we cannot have packer_src and packer_dst formats be the same because the packer gasket cannot perform conversion from `Float32` to `Float16_a`, so the packer_src format is `Float32` and packer_dst format is `Float16_a`. In wormhole architecture, the packer formats are decided specifically according to dest accumulation.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/helpers/include/data_format_inference.h
+++ b/tests/helpers/include/data_format_inference.h
@@ -1,0 +1,62 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+#include "tensix_types.h"
+
+#ifdef ARCH_WORMHOLE
+const bool is_blackhole = false;
+const bool is_wormhole  = true;
+#endif
+
+#ifdef ARCH_BLACKHOLE
+const bool is_blackhole = true;
+const bool is_wormhole  = false;
+#endif
+
+struct FormatConfig
+{
+    const uint32_t unpack_src;
+    const uint32_t unpack_dst;
+    const uint32_t pack_src;
+    const uint32_t pack_dst;
+};
+
+constexpr bool is_exponentB(uint32_t format)
+{
+    return (
+        format == static_cast<uint32_t>(DataFormat::Float16_b) || format == static_cast<uint32_t>(DataFormat::Bfp8_b) ||
+        format == static_cast<uint32_t>(DataFormat::Tf32));
+}
+
+constexpr bool is_format_combination_outlier(uint32_t input, uint32_t output, bool is_fp32_dest_acc_en)
+{
+    return (is_exponentB(input) && output == (uint32_t)DataFormat::Float16 && !is_fp32_dest_acc_en);
+}
+
+constexpr FormatConfig get_data_formats(uint32_t input, uint32_t output, bool is_fp32_dest_acc_en)
+{
+    uint32_t unpack_in  = input;
+    uint32_t unpack_out = input;
+    uint32_t pack_out   = output;
+    uint32_t pack_in;
+
+    if (input == (uint32_t)DataFormat::Float16 && output == (uint32_t)DataFormat::Bfp8_b && !is_fp32_dest_acc_en)
+    {
+        pack_in = static_cast<uint32_t>(DataFormat::Bfp8);
+    }
+    else if (is_wormhole && is_fp32_dest_acc_en && output == (uint32_t)DataFormat::Float16)
+    {
+        pack_in = static_cast<uint32_t>(DataFormat::Float32); // Gasket in wormhole cannot convert fp32 to fp16, and since dest accumulation turns on for
+                                                              // outlier cases we have fp32 in dest, so gasket cannot conver it to fp16, packer must do that
+    }
+    else if (is_format_combination_outlier(input, output, is_fp32_dest_acc_en))
+    {
+        pack_in = (is_wormhole) ? (uint32_t)DataFormat::Float32 : output;
+    }
+    else
+    {
+        pack_in = is_fp32_dest_acc_en ? output : input;
+    }
+
+    return {unpack_in, unpack_out, pack_in, pack_out};
+}
--- a/tests/helpers/include/params.h
+++ b/tests/helpers/include/params.h
@@ -10,6 +10,7 @@
 #include "ckernel_sfpu_log.h"
 #include "ckernel_sfpu_sqrt.h"
 #include "ckernel_sfpu_square.h"
+#include "data_format_inference.h"
 #include "llk_defs.h"
 #include "llk_sfpu_types.h"
 #include "tensix_types.h"
@@ -26,6 +27,20 @@ constexpr std::underlying_type_t<DataFormat> get_data_format(DataFormat format)
     return static_cast<std::underlying_type_t<DataFormat>>(format);
 }
 } // namespace
+
+constexpr bool dest_acc_en_input =
+#ifdef DEST_ACC
+    true;
+#else
+    false;
+#endif
+
+constexpr bool unpack_to_dest =
+#if defined(UNPACK_A_SRC_INT32) || defined(UNPACK_A_SRC_FLOAT32)
+    true;
+#else
+    false;
+#endif
 
 #define UNPACK_A_SRC_CASE(data_format) constexpr auto UNPACK_A_IN = get_data_format(DataFormat::data_format);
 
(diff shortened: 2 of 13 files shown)


tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifdef ARCH_BLACKHOLE
+const bool ARCH_WORMHOLE = true;
+#endif
+
+struct Formats {
+    const uint32_t unpack_src; 
+    const uint32_t unpack_dst;
+    const uint32_t pack_src;
+    const uint32_t pack_dst;
```
fvranicTT, line 15:
```suggestion
    DataFormat unpack_src; 
    DataFormat unpack_dst;
    DataFormat pack_src;
    DataFormat pack_dst;
```

tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,51 @@
+#include "tensix_types.h"
+
+#ifdef ARCH_WORMHOLE
+const bool ARCH_BLACKHOLE = false;
```
fvranicTT, line 4: It might be better to name it differently from the define. It will be easier later to read through the code (e.g. `is_blackhole` or similar)

tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,51 @@
...
+    const uint32_t unpack_dst;
+    const uint32_t pack_src;
+    const uint32_t pack_dst;
+};
+
+constexpr bool is_exponentB(uint32_t format){
+    return (format == static_cast<uint32_t>(DataFormat::Float16_b) || format == static_cast<uint32_t>(DataFormat::Bfp8_b) || format == static_cast<uint32_t>(DataFormat::Tf32));
+}
+
+constexpr bool format_combo_is_outlier(uint32_t input, uint32_t output, bool is_fp32_dest_acc_en){
```
fvranicTT, line 22:
```suggestion
constexpr bool format_combo_is_outlier(DataFormat input, DataFormat output, bool is_fp32_dest_acc_en){
```
Let's try to be more expressive with the `DataFormat` enum. That way, this function becomes:

```cpp
constexpr bool format_combo_is_outlier(DataFormat input, DataFormat output, bool is_fp32_dest_acc_en) {
    return is_exponentB(input) && output == DataFormat::Float16 && !is_fp32_dest_acc_en;
}
```

tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,51 @@
...
+    return (is_exponentB(input) && output  == (uint32_t)DataFormat::Float16 && !is_fp32_dest_acc_en);
+}
+
+constexpr Formats get_data_formats(uint32_t input, uint32_t output, bool is_fp32_dest_acc_en)
```
fvranicTT, line 26:
```cpp
constexpr Formats get_data_formats(DataFormat input, DataFormat output, bool is_fp32_dest_acc_en)
{
    DataFormat unpack_in = input;
    DataFormat unpack_out = input;
    DataFormat pack_out = output;
    DataFormat pack_in;

    if (input == DataFormat::Float16 && output == DataFormat::Bfp8_b && !is_fp32_dest_acc_en) {
        pack_in = DataFormat::Bfp8;
    } 
    else if (ARCH_WORMHOLE && is_fp32_dest_acc_en && output == DataFormat::Float16) {
        pack_in = DataFormat::Float32;
    } 
    else if (format_combo_is_outlier(input, output, is_fp32_dest_acc_en)) {
        pack_in = (ARCH_WORMHOLE) ? DataFormat::Float32 : output;
    } 
    else {
        pack_in = is_fp32_dest_acc_en ? output : input;
    }

    return {unpack_in, unpack_out, pack_in, pack_out};
}
```

tests/helpers/include/params.h
```diff
@@ -165,6 +179,20 @@ MATH_CASE(Bfp8_b)
 
 #undef MATH_CASE
 
+
+#if !defined(MATH_BFP8_B) && !defined(MATH_INT32) && !defined(MATH_FLOAT32) && !defined(MATH_FLOAT16) && !defined(MATH_FLOAT16_B)
+constexpr bool is_fp32_dest_acc_en = dest_acc || format_combo_is_outlier(UNPACK_A_IN, PACK_OUT, dest_acc);
+constexpr Formats pipeline_formats = get_data_formats(UNPACK_A_IN, PACK_OUT, dest_acc);
```
fvranicTT, line 185:
```suggestion
constexpr Formats pipeline_formats = get_data_formats(UNPACK_A_IN, PACK_OUT, is_fp32_dest_acc_en);
```
Shouldn't it be like this now?

tests/helpers/include/data_format_inference.h
fvranicTT: Also, can we find a better name for this method, like `is_format_outlier` or something similar?

tests/helpers/include/data_format_inference.h
skotaracTT: Has to be a constant because Format class is used in dramas to define constant across used as arguments to llk function parameters

tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifdef ARCH_WORMHOLE
+const bool ARCH_BLACKHOLE = false;
+#endif
+
+#ifdef ARCH_BLACKHOLE
+const bool ARCH_WORMHOLE = true;
+#endif
+
+struct Formats {
```
fvranicTT, line 11:
```suggestion
struct DataFormats {
```

tests/helpers/include/params.h
```diff
@@ -27,6 +28,19 @@ constexpr std::underlying_type_t<DataFormat> get_data_format(DataFormat format)
...
+const bool dest_acc = true;
+#else
+const bool dest_acc = false;
+#endif
+
+#if defined(UNPACK_A_SRC_INT32) || defined(UNPACK_A_SRC_FLOAT32) 
+const bool unpack_to_dest = true;
+#else
+const bool unpack_to_dest = false;
+#endif
```
fvranicTT, line 41:
```suggestion
constexpr bool dest_acc = 
#ifdef DEST_ACC
    true;
#else
    false;
#endif

constexpr bool unpack_to_dest = 
#if defined(UNPACK_A_SRC_INT32) || defined(UNPACK_A_SRC_FLOAT32)
    true;
#else
    false;
#endif
```

tests/helpers/include/data_format_inference.h
skotaracTT: Agreed!

tests/helpers/include/data_format_inference.h
skotaracTT: For now it's an outlier because the current combination doesn't work and there is no cohesive logic/explanation as to why. But we expect it to be resolved with HW team and then we will have no need for this function, at which point it will be removed. I can change the name of the function or add a comment explaining the situation.

tests/helpers/include/data_format_inference.h
skotaracTT: DataFormats might be confusing because we have a DataFormats class, so maybe something like FormatConfig? This would match with FormatConfig class input python and be quite intuitive

tests/helpers/include/params.h
skotaracTT: I understand your confusion...Some format combinations are not supported when dest accumulation is off, so we have to turn dest accumulation on. In order to avoid doing this check and changing every cpp test file, instead I define an intermediate value `dest_acc` that holds the flag for dest accumulation according to the python test setup. Then use this value and check if dest accumulation is off and the combination is an outlier which means dest accumulation needs to be turned on. 
```cpp
constexpr bool is_fp32_dest_acc_en = dest_acc || is_format_combination_outlier(UNPACK_A_IN, PACK_OUT, dest_acc);
```
Since `is_fp32_dest_acc_en` is conditionally defined and the argument for dest accumulation in llk function is passed in as a constant, so we can't redefine it, we must use `dest_acc` in order to define `is_fp32_dest_acc_en` once.

tests/helpers/include/params.h
skotaracTT: Works! Much cleaner too 👍

tests/helpers/include/data_format_inference.h
fvranicTT: Sounds like a plan.

tests/helpers/include/data_format_inference.h
fvranicTT: That would do.

tests/helpers/include/params.h
skotaracTT: Yes sure!

tests/helpers/include/data_format_inference.h
skotaracTT: Nice fix!👌

tests/helpers/include/data_format_inference.h
```diff
@@ -0,0 +1,62 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+#include "tensix_types.h"
+
+#ifdef ARCH_WORMHOLE
```
fvranicTT, line 6:
```cpp
#if defined(ARCH_WORMHOLE) && defined(ARCH_BLACKHOLE)
#error "Only one of ARCH_WORMHOLE or ARCH_BLACKHOLE can be defined"
#elif defined(ARCH_WORMHOLE)
constexpr bool is_blackhole = false;
constexpr bool is_wormhole  = true;
#elif defined(ARCH_BLACKHOLE)
constexpr bool is_blackhole = true;
constexpr bool is_wormhole  = false;
#else
#error "You must define either ARCH_WORMHOLE or ARCH_BLACKHOLE"
#endif
```

fvranicTT (approved):

tests/helpers/include/data_format_inference.h
skotaracTT: I will add this in next PR shortly

Issue #126, opened 2025-04-24
Add Input and Output format Parametrization to Test infrastructure


skotaracTT:
Adding ability to only pass input and output formats, instead of all formats or just one format being set to all formats in LLK pipeline. 
This feature goes hand in hand with Data Format Inference Model, without the model we cannot have only input and output parametrization.

PR #144, opened 2025-04-29
[Test Infra] Input Output Format Parametrization

skotaracTT:
### Ticket

https://github.com/tenstorrent/tt-llk/issues/126
### What's changed

Adding ability to pass only input and output formats, instead of all formats or just one format being set to all formats in LLK pipeline. This adds flexibility for user to define what they want in input and what format they want in output (in L1) so that they can work with their data in the desired formats further. Along with the Data Inference Model this feature works to let user define the formats they want and the inference model does the rest by setting the intermediate formats accordingly for the pipeline to work.

Note: This feature goes hand-in-hand with Data Format Inference Model, without this model we cannot have only input and output parametrization because the intermediate formats for unpacker, math and packer are not defined.
### Type of change

- [ ] Documentation update

### Checklist


tests/Makefile
```diff
@@ -38,7 +38,10 @@ OPTIONS_LINK=-fexceptions -Wl,-z,max-page-size=16 -Wl,-z,common-page-size=16 -no
 INCLUDES = -I../$(ARCH_LLK_ROOT)/llk_lib -I../$(ARCH_LLK_ROOT)/common/inc -I../$(ARCH_LLK_ROOT)/common/inc/sfpu -I../$(ARCH_LLK_ROOT)/hw_specific/inc
 INCLUDES += -Ifirmware/riscv/common -Ifirmware/riscv/$(CHIP_ARCH)/ -Isfpi/include -Ihelpers/include
 
-FORMAT_ARG := -D$(unpack_A_src) -D$(unpack_A_dst) -D$(unpack_B_src) -D$(unpack_B_dst) -D$(fpu) -D$(pack_src) -D$(pack_dst)
+FORMAT_ARG := -D$(unpack_A_src) -D$(pack_dst)
+ifdef fpu
+	FORMAT_ARG += -D$(unpack_A_dst) -D$(unpack_B_src) -D$(unpack_B_dst) -D$(fpu) -D$(pack_src)
```
fvranicTT, line 43:
```suggestion
	FORMAT_ARG += $(foreach var, unpack_A_dst unpack_B_src unpack_B_dst fpu pack_src, -D$($(var)))
endif
```
This is a nit. `$(foreach var, list, text)` iterates over each item in a list, makes it easier for later extension.

tests/sources/eltwise_unary_datacopy_test.cpp
```diff
@@ -106,4 +106,4 @@ void run_kernel()
     _llk_pack_dest_section_done_<DstSync::SyncFull, is_fp32_dest_acc_en>();
 }
 
-#endif
+#endif
```
fvranicTT, line 109: These cpp files don't need changes. It will reduce the number of files and make it easier to read.

tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -16,24 +16,21 @@ def generate_golden(operand1, format):
...
         ),
     ]
 )
 
+input_output_formats = [InputOutputFormat(DataFormat.Float32, DataFormat.Float32), InputOutputFormat(DataFormat.Float32, DataFormat.Float16), InputOutputFormat(DataFormat.Float32, DataFormat.Float16_b), InputOutputFormat(DataFormat.Float32, DataFormat.Bfp8_b),
+                        InputOutputFormat(DataFormat.Float16, DataFormat.Float32), InputOutputFormat(DataFormat.Float16, DataFormat.Float16), InputOutputFormat(DataFormat.Float16, DataFormat.Float16_b), InputOutputFormat(DataFormat.Float16, DataFormat.Bfp8_b),
+                        InputOutputFormat(DataFormat.Float16_b, DataFormat.Float32), InputOutputFormat(DataFormat.Float16_b, DataFormat.Float16), InputOutputFormat(DataFormat.Float16_b, DataFormat.Float16_b), InputOutputFormat(DataFormat.Float16_b, DataFormat.Bfp8_b),
+                        InputOutputFormat(DataFormat.Bfp8_b, DataFormat.Float32), InputOutputFormat(DataFormat.Bfp8_b, DataFormat.Float16), InputOutputFormat(DataFormat.Bfp8_b, DataFormat.Float16_b), InputOutputFormat(DataFormat.Bfp8_b, DataFormat.Bfp8_b),
+                        ]
+# input_output_formats = [InputOutputFormat(DataFormat.Float16, DataFormat.Bfp8_b)]
```
fvranicTT, line 33:
```suggestion
```
Commented out code should not be part of our codebase :)

tests/python_tests/test_eltwise_unary_datacopy.py
```diff
@@ -16,24 +16,21 @@ def generate_golden(operand1, format):
...
-            DataFormat.Int32,  # index 5 is for pack_dst
-            DataFormat.Float32,  # index 6 is for math format
+            DataFormat.Float16_b,  # index 2 is for pack_src (if src registers have same formats)
+            DataFormat.Bfp8_b,  # index 3 is for pack_dst
+            DataFormat.Float16_b,  # index 4 is for math format
         ),
     ]
 )
 
+input_output_formats = [InputOutputFormat(DataFormat.Float32, DataFormat.Float32), InputOutputFormat(DataFormat.Float32, DataFormat.Float16), InputOutputFormat(DataFormat.Float32, DataFormat.Float16_b), InputOutputFormat(DataFormat.Float32, DataFormat.Bfp8_b),
```
fvranicTT, line 28: This list isn't' used, correct? Should it be a part of a subsequent pull request?

tests/python_tests/helpers/format_config.py
```diff
@@ -114,7 +114,32 @@ def __init__(
                 )
             self.unpack_B_src = unpack_B_src
             self.unpack_B_dst = unpack_B_dst
+    
+    def get_output_format(self) -> DataFormat:
```
fvranicTT, line 118: could we turn these 2 into properties?
```python
@property
def output_format(self) -> DataFormat:
    return self.pack_dst

@property
def input_format(self) -> DataFormat:
    return self.unpack_A_src
```
I avoided using `get_` as it isn't idiomatic.

tests/Makefile
skotaracTT: Works like a charm 👍

tests/python_tests/helpers/format_arg_mapping.py
```diff
@@ -69,6 +69,22 @@
     DataFormat.Int32: "PACK_DST_INT32",
 }
 
+input_dict = {
```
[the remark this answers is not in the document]
skotaracTT: My mistake, I forgot to remove input_dict and output_dict they are not in use and not needed.

tests/python_tests/helpers/format_config.py
```diff
@@ -114,7 +114,32 @@ def __init__(
...
+    They are used for format inference model to infer the rest of the formats for the LLk pipeline, instead of the user.
+    """
+    input: DataFormat
+    output: DataFormat
+    
+    def __init__(self, input_format: DataFormat, output_format: DataFormat):
+        self.input = input_format
+        self.output = output_format
+    
+    def get_output_format(self) -> DataFormat:
```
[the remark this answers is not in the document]
skotaracTT: The reason I added functions for this is because I need two classes for different format arguments to the test parameter.
One is when we pass all 5-7 formats using `FormatConfig` class here the input is `unpack_A_src` and the output is `pack_dst`. However, when the user wishes to pass just input and output variables they will just need input in L1 and output in L1 so I cannot call `formats.unpack_A_src` for the `InputOutputFormat` class then I would have to go to every python files that wishes to get input and output of the format and add an if condition to check which class it is and call `.input` or `.unpack_A_src` accordingly. This way we call the same function on each class and the class internally accesses the correct variables without us having to check.

tests/python_tests/test_eltwise_unary_datacopy.py
skotaracTT: Erased, this is coming in a later PR

tests/python_tests/test_eltwise_unary_datacopy.py
skotaracTT: Yes, I will erase it an leave it for subsequent pull request

tests/sources/eltwise_unary_datacopy_test.cpp
skotaracTT: Agreed, this might have happened on accident.Pre-commit should be able to revert this added lines

tests/python_tests/helpers/format_config.py
skotaracTT: Agreed! I got rid of it 🤝

tests/python_tests/helpers/format_config.py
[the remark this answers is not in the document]
fvranicTT: This sounds like polymorphism. Thanks for explaining, I wasn't following deeply what goes on there.

tests/python_tests/helpers/device.py
```diff
@@ -117,21 +117,21 @@ def get_result_from_device(
     }
 
     # Handling "Bfp8_b" format separately with sfpu condition
-    if formats.pack_dst == DataFormat.Bfp8_b:
+    if formats.output_format() == DataFormat.Bfp8_b:
```
fvranicTT, line 120:
```suggestion
    if formats.output_format == DataFormat.Bfp8_b:
```
Properties can be accessed without (). It looks more like accessing a field, not calling a function.

tests/python_tests/helpers/format_config.py
```diff
@@ -115,6 +115,34 @@ def __init__(
...
+    """
+
+    input: DataFormat
+    output: DataFormat
+
+    def __init__(self, input_format: DataFormat, output_format: DataFormat):
+        self.input = input_format
+        self.output = output_format
+
+    def output_format(self) -> DataFormat:
```
fvranicTT, line 140:
```suggestion
    @property
    def output_format(self) -> DataFormat:
```

tests/python_tests/helpers/format_config.py
```diff
@@ -115,6 +115,34 @@ def __init__(
...
+        return self.output
+
+    def input_format(self) -> DataFormat:
+        return self.input
```
fvranicTT, line 144:
```suggestion
    @property
    def output_format(self) -> DataFormat:
        return self.output

    @property
    def input_format(self) -> DataFormat:
        return self.input
```

skotaracTT: Great!

tests/sources/eltwise_unary_datacopy_test.cpp
```diff
@@ -105,5 +105,4 @@ void run_kernel()
     _llk_pack_<DstSync::SyncFull, false, is_fp32_dest_acc_en>(0, L1_ADDRESS(buffer_Dest));
     _llk_pack_dest_section_done_<DstSync::SyncFull, is_fp32_dest_acc_en>();
 }
-
```
fvranicTT, line 108: you missed a spot :)

skotaracTT: In the next PR 🤝 pre-commit hooks passed for this run

PR #170, opened 2025-05-07
chore: update setup_testing_env.sh to display info in case of error

fvranicTT:
### Ticket
None

### Problem description
Currently, we are experiencing [random failures](https://github.com/tenstorrent/tt-llk/actions/runs/14880439156/job/41787137193?pr=169) caused by network issues when downloading the SFPI compiler. However, the error is misleading as it only appears during the pytest execution step. Instead, the failure should occur earlier, during the step where the SFPI compiler is being downloaded.

### What's changed
With this change, GH pipeline should stop if it fails to fetch the compiler or to unpack it. Pytests shouldn't even run.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/setup_testing_env.sh
+++ b/tests/setup_testing_env.sh
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
-# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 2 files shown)


tests/setup_testing_env.sh
```diff
@@ -3,14 +3,25 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+set -e  # Exit immediately if a command exits with a non-zero status
+set -o pipefail  # Fail if any command in a pipeline fails
+
```
skotaracTT, line 8: This is great!👏

skotaracTT (approved):

PR #180, opened 2025-05-08
feat: stop using globbing when importing utility methods

fvranicTT:
### Problem description
We've come across issues due to using globbing operator for utility imports. This was a relic of the original implementation that had to be tackled at one point in time.

### What's changed

1.  **Removed imports using globbing**
Globbing (from module import *) is bad because it hides dependencies, pollutes the namespace, and increases the risk of circular imports. 

2.  **Skip Decorators:**
Introduced `@skip_for_wormhole` and `@skip_for_blackhole` decorators.
These decorators dynamically check the CHIP_ARCH environment variable to skip tests incompatible with specific architectures. If you want to skip a certain test for a specific architecture, just decorate it with e.g. `@skip_for_wormhole`

3. **Introduced ChipArchitecture enum**
Introduced `ChipArchitecture` enum so we could stop using string comparison since it's error prone.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,11 @@ repos:
   hooks:
     -  id: autoflake
        args: [--remove-all-unused-imports, --in-place, --recursive, --exclude=__init__.py]
+- repo: https://github.com/PyCQA/pylint
+  rev: v3.0.0  # Use the latest version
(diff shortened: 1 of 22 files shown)


skotaracTT (approved):
Looks good

PR #181, opened 2025-05-09
Reorder semaphore writes and mailbox writes in  _llk_unpack_get_tile_

jbaumanTT:
### Ticket
[](https://github.com/tenstorrent/tt-metal/issues/19201)

### Problem description
We're seeing hangs in  `_llk_unpack_release_tile_` on blackhole in sdpa. They seem to be caused when a   `semaphore_get(semaphore::UNPACK_OPERAND_SYNC)` is called on TRISC1 and the semaphore is already 0, so the get is ignored. Later on, the semaphore is stuck at 1.

The mailbox_write in `_llk_unpack_get_tile_` signals TRISC1 to continue (in `_llk_math_get_tile_`), and very soon afterwards it executes semaphore_get in  `_llk_math_release_tile_` . If semaphore_post in `_llk_unpack_get_tile_` executes after the semaphore_get (due to an unlucky cache miss or similar) we'll reach that invalid state.

### What's changed
Move the semaphore_post before the `mailbox_write` in `_llk_unpack_get_tile_`, which ensures everything happens in order.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_common.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_common.h
@@ -40,14 +40,14 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
 
     if constexpr (mail2math)
     {
-        mailbox_write(ThreadId::MathThreadId, byte_address);
         semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        mailbox_write(ThreadId::MathThreadId, byte_address);
     }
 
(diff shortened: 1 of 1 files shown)


jbaumanTT:
I don't know if we want to copy these changes over to wormhole, since we haven't seen evidence of problems there.

fvranicTT:
It seems like @amahmudTT did just that in #185

jbaumanTT:
> It seems like @amahmudTT did just that in #185

Yeah, we can just go with his patch if that makes sense.

PR #185, opened 2025-05-09
Fix ordering of mailbox_write and semaphore_post in unpack_get_tile

amahmudTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/19201

### Problem description
The unpacker in _llk_unpack_get_tile function used to do mailbox_writes and then post (increment) a semaphore. It is possible for math and pack thread to read the mailboxes and try to get (decrement) a semaphore which is already zero, thus dropping the decrements. Then later on in _llk_unpack_release_tile the unpacker would wait for the semaphore to be zero, but since the decrements were dropped it would get stuck.

### What's changed
The order of doing the mailbox write and semaphore post has been changed, so the packer and math thread would only try to decrement the semaphore after the sempahore post since that happens before the mailbox write now which will happen before packer and math thread can decrement since their execution is depedent on the mailbox read.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_common.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_common.h
@@ -40,14 +40,14 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
 
     if constexpr (mail2math)
     {
-        mailbox_write(ThreadId::MathThreadId, byte_address);
         semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        mailbox_write(ThreadId::MathThreadId, byte_address);
     }
 
     if constexpr (mail2pack)
     {
-        mailbox_write(ThreadId::PackThreadId, byte_address);
         semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        mailbox_write(ThreadId::PackThreadId, byte_address);
     }
 
     *p_tile = byte_address;
--- a/tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
@@ -41,14 +41,14 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
 
     if constexpr (mail2math)
     {
-        mailbox_write(ThreadId::MathThreadId, byte_address);
         semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        mailbox_write(ThreadId::MathThreadId, byte_address);
     }
 
     if constexpr (mail2pack)
     {
-        mailbox_write(ThreadId::PackThreadId, byte_address);
         semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        mailbox_write(ThreadId::PackThreadId, byte_address);
     }
 
     *p_tile = byte_address;


amahmudTT:
In SDPA where the core where this issue happens is supposed to dictate some form of data flow for other cores, this would cause other cores to starve of data and eventually hang in different places due to instruction buffer getting filled up. @jbaumanTT and Marco debugged it and John came up with this fix and it seems that we do not see any hangs with SDPA as of now.

nvelickovicTT (approved):

pcawleyTT:
If I’m following along correctly, the crux of the problem is: prior to this PR, the following interleaving of events is possible between two threads:

|Step|Unpack thread|Math thread (ditto pack thread)|
|--:|---|---|
|1|Assuming semaphore is zero||
|2|Mailbox write||
|3||Mailbox read|
|4||Semaphore decrement|
|5|Semaphore increment||
|6|Wait for semaphore equal to 0||

> **Quick aside:** if this was code running on the host CPU, it would be easy to suggest why this interleaving could occur: the thread on the left suffers an interrupt, the OS decides to deschedule it for a bit, and during its downtime, the other thread runs. The baby RISCV cores don’t have interrupts nor an OS, so a different scenario needs suggesting - maybe the thread on the left suffered an icache miss and then extreme contention on L1 meant it took 100 cycles for the instruction fetch to complete, maybe the thread on the left had its branch predictor predict a branch where there was none and dealing with that mispredict cost some cycles, maybe the thread on the left was having a debugger attach to it and pause it. In the end, the exact cause doesn’t matter, but those are some plausible causes.

**Why is this interleaving problematic?** Because “semaphore decrement” isn’t `atomic { while (semaphore == 0) { wait; } semaphore -= 1; }` or even just `atomic{ semaphore -= 1 }`, but instead `atomic{ if (semaphore != 0) semaphore -= 1 }` , so if the semaphore is zero at step 4, it’ll stay at zero, then step 5 increments it, then step 6 waits forever.

I can see three potential solutions:
1. Ensure that this particular interleaving cannot happen.
2. Change how semaphores are used here, and initialise the semaphore to say `8`, so step 1 is assuming `semaphore == 8` and step 6 is waiting for `semaphore == 8`.
3. Don’t use semaphores here, and use a different synchronisation mechanism instead (for example a mailbox going in the other direction).

This PR takes solution number 1: swapping the order of the mailbox write and the semaphore increment in the source code _should_ make this interleaving impossible, as then the unpack thread should do the semaphore increment before the mailbox write, the math thread can’t complete its mailbox read until after the mailbox write, and then the math thread can’t do its semaphore decrement until after the mailbox read completes (by which point we know `semaphore != 0`, as the decrement happens after the increment). I say _should_, because you really need to get in to the details to upgrade _should_ to _certainly will_: the baby RISCV cores can, in certain scenarios, perform store instructions out of order - if there are Tensix threads executing Tensix instructions on the sync unit, then the RISCV semaphore increment could take a very long time to happen, and if that was the case, even if the machine code says `store for semaphore increment; store for mailbox write` then the two stores could still happen in the reverse (i.e. problematic) order. Given the context, it is probably OK, as there’s probably not enough contention on the sync unit to cause a problem. That said, once you’ve spent enough time chasing down race conditions of this form, then being 99.99% certain that things are now fixed isn’t good enough, as you know that the 0.01% might come back to bite you in the future.

Perhaps you can convince yourself that the 99.99% is in fact 100%, on the grounds that the Tensix sync unit isn't going to be doing much when the relevant code runs. If you can't convince yourself of that, then the alternative is to start being serious about [memory ordering](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/BabyRISCV/MemoryOrdering.md). Wormhole doesn't have a functional `fence` instruction, so at least there we can't appeal to the semantics of `fence` to get what we want. Instead I'd suggest this handy little collection of subtle functions for the baby RISCV cores:
```c
#include <stdint.h>

// If `x` is the result of loading from memory, placing `consume_discard(x)` somewhere
// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")

// `consume_discard_pair(x, y)` is like `consume_discard(x); consume_discard(y)` but
// does its work in a single instruction rather than two.
#define consume_discard_pair(x, y) __asm volatile("and x0, %0, %1" : : "r"((x)), "r"((y)) : "memory")

// This function always returns zero, but does so in a way which makes the CPU think that
// the value of x is required to come up with that zero. That zero can then be incorporated in
// to some other computation, and the CPU will think that said computation depends on x.
static inline __attribute__((always_inline))
uint32_t depend_as_zero(uint32_t x) {
  uint32_t result;
  __asm("andi %0, %1, 0" : "=r"(result) : "r"(x));
  return result;
}

// This function stores a value to memory, and then immediately reads it back. This initially
// looks pointless, but turns out to be very useful for memory ordering: the load result will
// not be available until the store has completed.
static inline __attribute__((always_inline))
uint32_t store_then_load(uint32_t* addr, uint32_t to_store) {
  uint32_t result;
  __asm volatile("sw %2, %1; lw %0, %1" : "=r"(result) : "m"(*addr), "r"(to_store));
  return result;
}
```

Given these guys, if you want to be confident that a semaphore increment happens before a mailbox write, then you've got a few options:
```c
// Option 1:
uint32_t sem_tmp = store_then_load(&pc_buf_base[PC_BUF_SEMAPHORE_BASE + semaphore::UNPACK_OPERAND_SYNC], 0);
consume_discard(sem_tmp);
mailbox_write(ThreadId::MathThreadId, byte_address);

// Option 2:
uint32_t sem_tmp = store_then_load(&pc_buf_base[PC_BUF_SEMAPHORE_BASE + semaphore::UNPACK_OPERAND_SYNC], 0);
mailbox_write(ThreadId::MathThreadId, byte_address | depend_as_zero(sem_tmp));

// Variant on option 1 when there are two increments of the same semaphore:
if constexpr (mail2math || mail2pack) {
  if constexpr (mail2math && mail2pack) semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
  uint32_t sem_tmp = store_then_load(&pc_buf_base[PC_BUF_SEMAPHORE_BASE + 
  semaphore::UNPACK_OPERAND_SYNC], 0);
  consume_discard(sem_tmp);
  if constexpr (mail2math) mailbox_write(ThreadId::MathThreadId, byte_address);
  if constexpr (mail2pack) mailbox_write(ThreadId::PackThreadId, byte_address);
}
```

On Blackhole you might instead be able to just throw down an appropriate `fence` instruction.

nvelickovicTT:
> even if the machine code says `store for semaphore increment; store for mailbox write` then the two stores could still happen in the reverse (i.e. problematic) order.

@pcawleyTT I think even if this happens it should still be fine, since the `semaphore increment` and `semaphore decrement `will execute in the right order, ensured by the fact that they are executed by the same EXU which can only execute in-order.
And the hypothesis of the hang is that these got inverted and caused silent miss of the decrement.

pcawleyTT:
> @pcawleyTT I think even if this happens it should still be fine, since the `semaphore increment` and `semaphore decrement `will execute in the right order, ensured by the fact that they are executed by the same EXU and are in instruction buffer which can only execute in-order.

I disagree with this statement, as these are semaphore manipulations performed via the RISCV MMIO interface, rather than semaphore manipulations performed via Tensix instructions (and hence there's no instruction buffer in play). If they _were_ Tensix instructions, then there'd still be no guarantee, as the increment and decrement are performed by different threads, and the in-order properties of an instruction buffer only apply to each thread individually.

For reference, the sync exu is arbitrating six in-order input streams:
1. Tensix instructions from thread 0
2. Semaphore MMIO from TRISC 0
3. Tensix instructions from thread 1
4. Semaphore MMIO from TRISC 1
5. Tensix instructions from thread 2
6. Semaphore MMIO from TRISC 2

The arbitration scheme is:
- Fair arbitration of:
  - Unfair arbitration of:
    - Tensix instructions from thread 0 (always have priority)
    - Semaphore MMIO from TRISC 0
  - Unfair arbitration of:
    - Tensix instructions from thread 1 (always have priority)
    - Semaphore MMIO from TRISC 1
  - Unfair arbitration of:
    - Tensix instructions from thread 2 (always have priority)
    - Semaphore MMIO from TRISC 2

Due to the unfair arbitration, semaphore MMIO can be starved.

nvelickovicTT:
Ahh I see. Thanks for the clarifications, that makes sense.
It seems to me, if we stick with semaphores approach, the only 100% solution is to add wait on zero before decrementing, and wait on max before incrementing to the synchronization mechanism.
Otherwise, it's just a different degree of probability for things to go wrong.

pcawleyTT:
> the only 100% solution is to add wait on zero before decrementing

Alas, that's not a fix in this case, as there are two threads both doing a decrement on the same semaphore, so you can get this interleaving on the decrement threads:
|Step|Math thread|Pack thread|
|---|---|---|
|1|`while (semaphore == 0) wait;`||
|2||`while (semaphore == 0) wait;`|
|3||decrement semaphore|
|4|decrement semaphore||

amahmudTT:
All BH post commits pass. 
https://github.com/tenstorrent/tt-metal/actions/runs/14973991881

Issue #177, opened 2025-05-08
Dealing with files shared between tt-llk and tt-metal


fvranicTT:
Currently, we have several header files that reside in the `tt-metal` repository but are required for the LLK test infrastructure to compile and execute:
1.  `cfg_defines.h`
2. `dev_mem_map.h`
3. `tensix.h`
4. `tensix_types.h`
5. `ckernel_sfpu_converter.h`

These files primarily contain definitions (e.g. memory sizes, register addresses, etc.). Until now, copies of these files were stored in the `tt-llk` repository to break the dependency. However, since they are frequently updated in the `tt-metal` repository, we need to continuously track these changes to keep them in sync. This approach is not sustainable.

Since these files are essential and cannot be removed, an ideal solution would be to store them in a centralized location (e.g., a separate repository) that all other repositories could access. However, this would complicate matters for existing users (e.g., metal, umd).

The proposed solution is to download these files from `tt-metal` dynamically whenever LLK tests are run. This approach is minimally intrusive, as it only affects our repository and does not impact others. The main drawback is that we may encounter sudden CI failures if changes are made to these files in `tt-metal` and we are not prepared for them. Also, later, if we encounter issues with stack/code size, we could come back to the previous approach, as we would need to have our version of the above, suitable for testing. But I'd expect it be just one file.

PR #189, opened 2025-05-12
fix: remove hardware-specific header files shared with tt-metal

fvranicTT:
### Ticket
#177

### Problem description
Currently, we have several header files that reside in the tt-metal repository but are required for the LLK test infrastructure to compile and execute. These files primarily contain definitions (e.g. memory sizes, register addresses, etc.). Until now, copies of these files were stored in the tt-llk repository to break the dependency. However, since they are frequently updated in the tt-metal repository, we need to continuously track these changes to keep them in sync. This approach is not sustainable.

### What's changed
The proposed solution is to download these files from tt-metal dynamically whenever LLK tests are run. This approach is minimally intrusive, as it only affects our repository and does not impact others.
The change in this PR does just that: it downloades the required files on the fly, just in time for tests to run.

### Type of change

- [ ] Documentation update


lpremovicTT:
Should we own `ckernel_sfpu_converter.h` (in `<ARCH>/common/inc/sfpu`) and just call it from that file in metal, like with `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_dropout.h`?

fvranicTT:
> Should we own `ckernel_sfpu_converter.h` (in `<ARCH>/common/inc/sfpu`) and just call it from that file in metal, like with `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_dropout.h`?

Yes, we should.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_converter.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include <cstdint>
+
+namespace ckernel::sfpu
+{
+
+class Converter
+{
+public:
+    static float to_float(std::uint32_t value)
```
sstanisicTT, line 15: Might want to rename this for clarity while at it, because it doesn't convert, but reinterpret the value as float (I suggest `as_float` as better alternative).

lpremovicTT (approved):

PR #194, opened 2025-05-13
Introduced strict ordering in unpack_get_tile

amahmudTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/19201

### Problem description
The llk_unpack_get_tile is still prone to memory reording where the mailbox write can get executed before the semaphore post, leading to the race condition discussed in the issue. 

### What's changed
Modified the code to introduce reads and pseudo usage after store and load respectively, so the former get completed and eventually mailbox write happens after semaphore post. 

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_common.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_common.h
@@ -19,6 +19,20 @@
 using namespace ckernel;
 using namespace ckernel::unpacker;
 
+// If `x` is the result of loading from memory, placing `consume_discard(x)` somewhere
+// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
+#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")
+
+// This function stores a value to memory, and then immediately reads it back.
+// The load result will not be available until the store has completed.
+// This will make sure any subsequent instruction will see the store as complete.
+static inline __attribute__((always_inline)) uint32_t store_then_load(volatile uint32_t *addr, uint32_t to_store)
+{
+    uint32_t result;
+    __asm volatile("sw %2, %1; lw %0, %1" : "=r"(result) : "m"(*addr), "r"(to_store));
+    return result;
+}
+
 void _llk_zero_buffer_(const std::uint32_t base_address, const std::uint32_t size)
 {
     TT_SETDMAREG(0, 0, 0, LO_16(p_gpr_unpack::OPERAND_OFFSET_ADDR));
@@ -38,16 +52,25 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
 {
     std::uint32_t byte_address = (address) << 4;
 
-    if constexpr (mail2math)
+    if constexpr (mail2math || mail2pack)
     {
-        semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
-        mailbox_write(ThreadId::MathThreadId, byte_address);
-    }
+        if constexpr (mail2math && mail2pack)
+        {
+            semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
+        }
 
-    if constexpr (mail2pack)
-    {
-        semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
-        mailbox_write(ThreadId::PackThreadId, byte_address);
+        uint32_t sem_tmp = store_then_load(&pc_buf_base[PC_BUF_SEMAPHORE_BASE + semaphore::UNPACK_OPERAND_SYNC], 0);
+        consume_discard(sem_tmp);
+
+        if constexpr (mail2math)
+        {
+            mailbox_write(ThreadId::MathThreadId, byte_address);
+        }
+
+        if constexpr (mail2pack)
+        {
+            mailbox_write(ThreadId::PackThreadId, byte_address);
+        }
     }
 
     *p_tile = byte_address;
--- a/tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
@@ -20,6 +20,20 @@
 using namespace ckernel;
 using namespace ckernel::unpacker;
 
+// If `x` is the result of loading from memory, placing `consume_discard(x)` somewhere
+// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
+#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")
+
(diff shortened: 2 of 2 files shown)


tt_llk_blackhole/llk_lib/llk_unpack_common.h
```diff
@@ -19,6 +19,20 @@
 using namespace ckernel;
 using namespace ckernel::unpacker;
 
+// If `x` is the result of loading from memory, placing `consume_discard(x)` somewhere
+// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
+#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")
```
rtawfik01, line 24: Should the consume_discard and store_then_load be placed in the same function? Will there be a chance that we want these 2 separately? won't we always want the load to complete?

amahmudTT: Here we could put them in the same function but if we end up using them in the future we could have separate usages. Here the load is redundant and we want the redundant load to complete , but there could be cases where we have a valid load and no store there we would only need consume_discard and not the store_and_load.

fvranicTT: Isnt x0 on RISCV always 0?
Wouldn't `#define consume_discard(x) __asm volatile("" : : "r"(x) : "memory")` do the same?

amahmudTT: We want to give the idea that this is being used, so that it does not discard the load.  I am not 100% familiar with the format of the riscv inline assembly, but the code you wrote does not have any op or argument, or probably it is how git is showing what you typed.

tt_llk_blackhole/llk_lib/llk_unpack_common.h
```diff
@@ -38,16 +52,25 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
...
     std::uint32_t byte_address = (address) << 4;
 
-    if constexpr (mail2math)
-    {
-        semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
-        mailbox_write(ThreadId::MathThreadId, byte_address);
-    }
-
-    if constexpr (mail2pack)
+    if constexpr (mail2math || mail2pack)
```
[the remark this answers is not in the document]
amahmudTT: I am 50-50 about this, so prefer to keep it as it is.

tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
```diff
@@ -20,6 +20,20 @@
...
+// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
+#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")
+
+// This function stores a value to memory, and then immediately reads it back.
+// The load result will not be available until the store has completed.
+// This will make sure any subsequent instruction will see the store as complete.
+static inline __attribute__((always_inline)) uint32_t store_then_load(uint32_t *addr, uint32_t to_store)
+{
+    uint32_t result;
+    __asm volatile("sw %2, %1; lw %0, %1" : "=r"(result) : "m"(*addr), "r"(to_store));
```
fvranicTT, line 33: Do we need "memory" clobber here?

amahmudTT: We want the semaphore value to be updated and then read it back so that the compiler makes sure that the value is indeed updated before following instructions  are executed. Otherwise we might have the same race condition we were trying to avoid.

rtawfik01 (approved):

tt_llk_blackhole/llk_lib/llk_unpack_common.h
```diff
@@ -19,6 +19,20 @@
...
 using namespace ckernel::unpacker;
 
+// If `x` is the result of loading from memory, placing `consume_discard(x)` somewhere
+// will ensure that code after `consume_discard(x)` doesn't start until the load is complete.
+#define consume_discard(x) __asm volatile("andi x0, %0, 0" : : "r"((x)) : "memory")
+
+// This function stores a value to memory, and then immediately reads it back.
+// The load result will not be available until the store has completed.
+// This will make sure any subsequent instruction will see the store as complete.
+static inline __attribute__((always_inline)) uint32_t store_then_load(uint32_t *addr, uint32_t to_store)
```
fvranicTT, line 29: shouldn't addr be marked as volatile?
Also, could we use __atomic_store_n and __atomic_load_n?
```cpp
static inline __attribute__((always_inline)) uint32_t store_then_load(volatile uint32_t *addr, uint32_t to_store)
{
    __atomic_store_n(addr, to_store, __ATOMIC_SEQ_CST);
    return __atomic_load_n(addr, __ATOMIC_SEQ_CST);
}
```

amahmudTT: In general  compiler does not touch inline assembly, and line 32, uses the volatile keyword already.

ncvetkovicTT (approved):

tt_llk_blackhole/llk_lib/llk_unpack_common.h
```diff
@@ -38,16 +52,25 @@ inline void _llk_unpack_get_tile_(std::uint32_t address, std::uint32_t *p_tile)
...
-    }
-
-    if constexpr (mail2pack)
+    if constexpr (mail2math || mail2pack)
     {
-        semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
-        mailbox_write(ThreadId::PackThreadId, byte_address);
+        if constexpr (mail2math && mail2pack)
+        {
+            semaphore_post(semaphore::UNPACK_OPERAND_SYNC);
```
nvelickovicTT, line 59: I might be missing something here, but in the previous code `semaphore_post` was called twice in most cases because both `mail2math` and `mail2pack` were true. Here, it will be called once. Is that reflected also on the receving end when decrementing semaphore?

amahmudTT:
All Blackhole post commits pass https://github.com/tenstorrent/tt-metal/actions/runs/15028499135
All post commits pass https://github.com/tenstorrent/tt-metal/actions/runs/15028494570

tt_llk_blackhole/llk_lib/llk_unpack_common.h
nvelickovicTT: I see. That is not really obvious in that part of the code. Can you add a comment to describe what is going on and why is it needed.

amahmudTT: https://github.com/tenstorrent/tt-llk/pull/199

PR #203, opened 2025-05-15
[Test Infra] Added test to fuse matmul with unpack tilize

skotaracTT:
### Ticket

https://github.com/orgs/tenstorrent/projects/123?pane=issue&itemId=109488923&issue=tenstorrent%7Ctt-llk%7C173
### Problem description

Adding Matmul test coverage by testing matmul with unpack_tilize. This test cannot fuse this in one llk pipeline. We must first unpack tilize the input tensors and then pack them into L1 and unpack them again to run a matmul on them. 
Needed to incorporate synchronization with `t6_semaphore`, as well as reconfiguring data formats for the next unpack, math and pack kernels in the case that we want to alter our formats when we execute the matmul in the second llk pipeline run. In packer code block, reconfiguring the data formats also sets packer strides.
- In this specific case in the first pipeline run we set the packer strides to pack in tilized format, but in the second run we don't pack in tilized format so this is set properly in the `set_packer_strides` call which is invoked in the `_llk_pack_reconfig_data_format_` function we call when we pack the second time.

I also plan to re-init for next unpack kernel but for that I need to bring in `llk_unpack_tilize_uninit` from metal into our llk library, this is in another PR; once it is merged I will add it in this test file. 
- For the time being the test runs and passes but this is a part of good calling conventions and llk programming model

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/test_matmul_unpack_tilize.py
+++ b/tests/python_tests/test_matmul_unpack_tilize.py
@@ -4,24 +4,24 @@
 import pytest
 import torch
 
-from helpers.format_config import DataFormat
-from helpers.format_arg_mapping import format_dict, MathFidelity, DestAccumulation
-from helpers.param_config import (
-    clean_params,
-    generate_param_ids,
-    generate_params,
-    input_output_formats,
-)
 from helpers.device import (
     collect_results,
     run_elf_files,
     wait_for_tensix_operations_finished,
     write_stimuli_to_l1,
 )
+from helpers.format_arg_mapping import DestAccumulation, MathFidelity, format_dict
+from helpers.format_config import DataFormat
+from helpers.param_config import (
+    clean_params,
+    generate_param_ids,
+    generate_params,
+    input_output_formats,
+)
 from helpers.stimuli_generator import generate_stimuli
 from helpers.test_config import generate_make_command
-from helpers.utils import compare_pcc, run_shell_command
 from helpers.tilize_untilize import tilize
+from helpers.utils import compare_pcc, run_shell_command
 
 
 def generate_golden(operand1, operand2, data_format, math_fidelity):
--- a/tests/python_tests/test_matmul_unpack_tilize.py
+++ b/tests/python_tests/test_matmul_unpack_tilize.py
@@ -1,0 +1,133 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+
+from helpers.format_config import DataFormat
+from helpers.format_arg_mapping import format_dict, MathFidelity, DestAccumulation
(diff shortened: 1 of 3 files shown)


fvranicTT (approved):

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
```diff
@@ -185,7 +185,7 @@ inline void _llk_unpack_tilize_(
 #endif
 }
 
-inline void _llk_unpack_tilize_uninit_(const std::uint32_t num_faces, const std::uint32_t face_r_dim = FACE_R_DIM, const std::uint32_t unpack_dst_format = 0)
+inline void _llk_unpack_tilize_uninit_(const std::uint32_t unpack_dst_format, const std::uint32_t num_faces = 4, const std::uint32_t face_r_dim = FACE_R_DIM)
```
fvranicTT, line 188: ?

skotaracTT: I just added this as a place holder before my PR to check if al the tests run now. The new PR I made in main to add these functions will be replaced with this once I rebase

fvranicTT: Still seeing this diff. Should it disappear, right?

tests/python_tests/test_matmul_unpack_tilize.py
```diff
@@ -0,0 +1,143 @@
...
+        res_from_L1,
+        dtype=(
+            format_dict[formats.output_format]
+            if formats.output_format in [DataFormat.Float16, DataFormat.Float16_b]
+            else torch.bfloat16
+        ),
+    )
+
+    if formats.output_format in [DataFormat.Float16_b, DataFormat.Float16]:
+        atol = 0.1
```
fvranicTT, line 131: can we make this a bit tighter?

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
skotaracTT: The difference is fine now, because this is how it is one main. I've rebased

fvranicTT: This is what we have on main:

https://github.com/tenstorrent/tt-llk/blob/1cbbc49c7bbcf9c2989bfaeadbc295c09d2fa332/tt_llk_blackhole/llk_lib/llk_unpack_tilize.h#L188C1-L189C2

```cpp
inline void _llk_unpack_tilize_uninit_(const std::uint32_t num_faces, const std::uint32_t face_r_dim = FACE_R_DIM, const std::uint32_t unpack_dst_format = 0)
```

fvranicTT: So it's not the same as in main ;)

PR #222, opened 2025-05-20
Activating Data Format Inference model with Input Output format parametrization Flexibility for same formats 

skotaracTT:
### Problem description

User should run all tests using input-output parametrization in order to utilize data format inference model. This is very useful because all input-output format combinations are supported by the data format inference model. If we hard code the combinations we might make a mistake as some hard-coded combinations with certain flags such as dest_Acc are not supported on certain architecture, and the data format inference model takes care of this in order to support all input-output combinations and increase coverage.

Fused tests that incorporate 2 or more llk pipeline runs currently don't utilize data format inference model because all the formats are the same (in order to reuse output of first call as input to second llk pipeline run ...) 
However if all formats are the same this isn't always supported by hardware for example if all formats are float16_an and dest is in float32 mode, on bh this works but fails on wh. The data format inference model accounts for this and sets the pack input format to float32.

By running on all tests with input-output format we activate the data format inference model to create combinations for each input-output combination. This increases coverage and we don't have to skip test cases that work with minor tweaking.

### What's changed

Added flag to input-output formats sweep when we want input and output format to be the same.
Run tests using this flag when input and output need to be the same so that the output is reused for fused tests.
### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/helpers/param_config.py
+++ b/tests/python_tests/helpers/param_config.py
@@ -289,7 +289,7 @@ def generate_param_ids(included_params, all_params: List[tuple]) -> List[str]:
 
 
 def input_output_formats(
-    formats: List[DataFormat], same: Optional[bool] = False
+    formats: List[DataFormat], same: bool = False
 ) -> List[InputOutputFormat]:
     """
     Generates a list of InputOutputFormat instances based on the given formats.
--- a/tests/python_tests/helpers/param_config.py
+++ b/tests/python_tests/helpers/param_config.py
@@ -288,7 +288,9 @@ def generate_param_ids(included_params, all_params: List[tuple]) -> List[str]:
     return [format_combination(comb) for comb in all_params if comb[0] is not None]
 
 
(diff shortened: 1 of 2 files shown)


tests/python_tests/helpers/param_config.py
```diff
@@ -288,7 +288,7 @@ def format_combination(comb: tuple) -> str:
     return [format_combination(comb) for comb in all_params if comb[0] is not None]
 
 
-def input_output_formats(formats: List[DataFormat]) -> List[InputOutputFormat]:
+def input_output_formats(formats: List[DataFormat], same: Optional[bool]=False) -> List[InputOutputFormat]:
```
fvranicTT, line 291:
```suggestion
def input_output_formats(formats: List[DataFormat], same: bool=False) -> List[InputOutputFormat]:
```
No point of having optional with default value :)

fvranicTT (approved):

tests/python_tests/helpers/param_config.py
skotaracTT: Completely agree! 👍😀

PR #230, opened 2025-05-22
Rewrite {recip,rsqrt,sqrt} LLKs.

jasondavies:
See:

- tenstorrent/tt-metal#21622

The original implementations used poor initial guesses for Newton-Raphson, resulting in poor results even when using 3 iterations.

These new implementations use state-of-the-art constants for the initial guesses, along with algorithmic improvements.

The new implementations have ~1 ulps error for either approximate mode (10 bits) or non-approximate mode (~23 bits).

The sqrt LLK now also has support for optional rsqrt (reciprocal sqrt), which uses the same number of multiplies as sqrt.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_binary.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_binary.h
@@ -144,7 +144,7 @@ inline void _calculate_sfpu_binary_(const uint dst_offset)
             }
             v_else
             {
-                result = in0 * sfpi::setsgn(_sfpu_reciprocal_<4>(in1), in1);
+                result = in0 * sfpi::setsgn(_sfpu_reciprocal_(in1), in1);
             }
             v_endif;
         }
@@ -165,9 +165,14 @@ inline void _calculate_sfpu_binary_(const uint dst_offset)
 template <bool APPROXIMATION_MODE /*unused*/, BinaryOp BINOP>
 inline void _sfpu_binary_init_()
 {
-    if constexpr (BINOP == BinaryOp::DIV || BINOP == BinaryOp::POW)
+    if constexpr (BINOP == BinaryOp::DIV)
     {
-        _init_reciprocal_<APPROXIMATION_MODE>();
+        _init_reciprocal_<false>();
+    }
+    if constexpr (BINOP == BinaryOp::POW)
+    {
+        // note: calls _init_reciprocal_
+        _init_exponential_<false>();
     }
 }
 
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_exp.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_exp.h
@@ -25,7 +25,7 @@ sfpi_inline sfpi::vFloat _sfpu_exp_(sfpi::vFloat val)
     v_endif;
 
     // Run series in Horner form
-    sfpi::vFloat tmp = val * sfpi::vConst0p8373 + sfpi::s2vFloat16b(0.863281);
+    sfpi::vFloat tmp = val * sfpi::vConst0p8373 + sfpi::vConstFloatPrgm2;
     val              = val * tmp + sfpi::vConst1;
 
     v_if (exp >= 0)
@@ -360,8 +360,7 @@ inline void _init_exponential_()
     }
     else
     {
-        sfpi::vConstFloatPrgm0 = 1.442695f; // ln2_recip
-        sfpi::vConstFloatPrgm1 = 2.0f;
+        _init_reciprocal_<APPROXIMATION_MODE>();
         sfpi::vConstFloatPrgm2 = 0.863281f;
     }
 }
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_recip.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_recip.h
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: © 2025 Jason Davies <jason@jasondavies.com>
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -11,61 +12,37 @@ namespace ckernel
 namespace sfpu
 {
 
-template <int max_iter = 3>
+// See: Cezary J. Walczyk, Leonid V. Moroz, Volodymyr Samotyy, and Jan L. Cieśliński.
+// Optimal Approximation of the 1/x Function using Chebyshev Polynomials and Magic Constants.
+// https://doi.org/10.1145/3708472
+
+template <bool APPROXIMATE = false>
 sfpi_inline sfpi::vFloat _sfpu_reciprocal_(const sfpi::vFloat in)
 {
-    // Force sign to 1 (make number negative)
-    sfpi::vFloat val = sfpi::setsgn(in, 1);
+    sfpi::vFloat negative_x = -in;
+    sfpi::vFloat y = sfpi::reinterpret<sfpi::vFloat>(sfpi::vConstIntPrgm0 - sfpi::reinterpret<sfpi::vInt>(in));
+    sfpi::vFloat t = sfpi::vConstFloatPrgm2 + negative_x * y;
+    y = y * sfpi::vConstFloatPrgm1;
+    y = y * t;
 
-    val = setexp(val, 126); // Set exponent to 126 to make the number in 0.5-1
-    // Use 1.44 as first guess at x, ideal value would be 1.33.
-    // Grayskull has hardwired 1.44 and uses it to avoid a load.
-    // We use it here for consistency.
-    sfpi::vFloat vConstLn2Recip = sfpi::vConstFloatPrgm0;
-    sfpi::vFloat two            = sfpi::vConstFloatPrgm1;
-    sfpi::vFloat result         = vConstLn2Recip * (val * vConstLn2Recip + two);
-
-    for (int s_iter = 0; s_iter < (max_iter - 1); s_iter++)
+    if constexpr (!APPROXIMATE)
     {
-        result = result * (val * result + two);
+        // 2nd iteration of Newton-Raphson
+        t = y * negative_x + sfpi::vConst1;
+        y = y * t + y;
     }
 
-    sfpi::vInt orig_exp = exexp(in);
-    sfpi::vInt new_exp  = exexp(result);
-
-    // "Subtract" exponents, and re-bias.
-    // Execute: -1 - exp, then exp += 127
-    new_exp -= orig_exp;
-    new_exp += 126;
-
-    v_if (new_exp < 0)
-    {
-        // If rebiased exponent is negative, we need to saturate at 0.
-        // This means the initial number was too big so reciprocal result should be 0
-        result  = 0.0F;
-        new_exp = 0;
-    }
-    v_endif;
-
-    // Set newly denormalized exponent to result exponent field
-    return setexp(result, new_exp);
+    return y;
 }
 
-template <bool APPROXIMATION_MODE, int ITERATIONS, bool is_fp32_dest_acc_en = true>
+template <bool APPROXIMATION_MODE, bool is_fp32_dest_acc_en = true>
 inline void _calculate_reciprocal_(const int iterations)
 {
 #pragma GCC unroll 8
     for (int d = 0; d < iterations; d++)
     {
         sfpi::vFloat in  = sfpi::dst_reg[0];
-        sfpi::vFloat out = _sfpu_reciprocal_<APPROXIMATION_MODE ? 2 : 3>(in);
-
-        v_if (in < 0.0F)
-        {
-            // Invert sign on calculated value if CC=1 (number is negative)
-            out = -out;
-        }
-        v_endif;
+        sfpi::vFloat out = _sfpu_reciprocal_<APPROXIMATION_MODE>(in);
 
         if constexpr (is_fp32_dest_acc_en || APPROXIMATION_MODE)
         {
@@ -83,8 +60,9 @@ inline void _calculate_reciprocal_(const int iterations)
 template <bool APPROXIMATION_MODE>
 inline void _init_reciprocal_()
 {
-    sfpi::vConstFloatPrgm0 = 1.442695f; // ln2_recip
-    sfpi::vConstFloatPrgm1 = 2.0f;
+    sfpi::vConstIntPrgm0 = 0x7eb504f3;
+    sfpi::vConstFloatPrgm1 = 1.94090888923f;
+    sfpi::vConstFloatPrgm2 = 1.43566017178f;
 }
 
 } // namespace sfpu
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_sqrt.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_sqrt.h
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: © 2025 Jason Davies <jason@jasondavies.com>
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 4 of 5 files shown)


tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_recip.h
```diff
@@ -11,61 +12,37 @@ namespace ckernel
...
-        v_if (in < 0.0F)
-        {
-            // Invert sign on calculated value if CC=1 (number is negative)
-            out = -out;
-        }
-        v_endif;
+        sfpi::vFloat out = _sfpu_reciprocal_<APPROXIMATION_MODE>(in);
 
         if constexpr (is_fp32_dest_acc_en || APPROXIMATION_MODE)
         {
```
jasondavies, line 48: Can anyone explain why this is here?

jasondavies: By "this", I mean the line below (that I can't seem to add a comment to), that says:

```cpp
sfpi::dst_reg[0] = sfpi::reinterpret<sfpi::vFloat>(float_to_fp16b(out, 0));
```

This happens when is_fp32_dest_acc_en=false and APPROXIMATION_MODE=false.

jasondavies: See [here](https://github.com/tenstorrent/tt-llk/commit/9f00b53fe5583dc8946c4e66c89b127e37d06584#r157770862) for where it got added by @rdjogoTT.

rdjogoTT: fp16b formats in Dest were having accuracy issues because the recip is computed in fp32 in the LREGs, but when we store datums back to Dest from LREGs it truncates the lower bits. This was added to do rounding before storing values back to Dest, and helped with the accuracy problem. We do have an internal issue open to potentially look at adding this to other places as well, it's just low priority right now.

jasondavies: In an ideal world it would probably be good to support rounding to both bf16 and fp16, as this assumes bf16 right now.

fvranicTT:
@jasondavies Can you try adding the tests following the same approach as in https://github.com/tenstorrent/tt-llk/pull/101?
See https://github.com/tenstorrent/tt-llk/blob/main/tests/README.md on how to run the tests. One note: at the moment automatic detection of cards is broken, I'm looking into it, but it should work if you `export CHIP_ARCH=wormhole`.
These tests are checking pcc, not ideal thing, but should be enough for now. We plan to start checking the ULPs, but we're not there yet.

jasondavies:
I'm confused why Sin/Cos tests are failing as I haven't touched them:

```
>           assert torch.isclose(
                golden_tensor[i], res_tensor[i], rtol=rtol, atol=atol
            ), f"Failed at index {i} with values {golden[i]} and {res_from_L1[i]}"
E           AssertionError: Failed at index 0 with values 0.937653277514828 and 131008.0
E           assert tensor(False)
E            +  where tensor(False) = <built-in method isclose of type object at 0x7f2937f86bc0>(tensor(0.94, dtype=torch.bfloat16), tensor(131072., dtype=torch.bfloat16), rtol=0.1, atol=0.05)
E            +    where <built-in method isclose of type object at 0x7f2937f86bc0> = torch.isclose

test_eltwise_unary_sfpu.py:169: AssertionError
```

fvranicTT:
> I'm confused why Sin/Cos tests are failing as I haven't touched them:
> 
> ```
> >           assert torch.isclose(
>                 golden_tensor[i], res_tensor[i], rtol=rtol, atol=atol
>             ), f"Failed at index {i} with values {golden[i]} and {res_from_L1[i]}"
> E           AssertionError: Failed at index 0 with values 0.937653277514828 and 131008.0
> E           assert tensor(False)
> E            +  where tensor(False) = <built-in method isclose of type object at 0x7f2937f86bc0>(tensor(0.94, dtype=torch.bfloat16), tensor(131072., dtype=torch.bfloat16), rtol=0.1, atol=0.05)
> E            +    where <built-in method isclose of type object at 0x7f2937f86bc0> = torch.isclose
> 
> test_eltwise_unary_sfpu.py:169: AssertionError
> ```
> 
> Edit: looks like they break on main too, so nothing to do with me!

That’s odd. They were passing yesterday.

fvranicTT:
Let's disable it for now. I can take care of that.

jasondavies:
I've only disabled the Cos tests, which were failing for some reason.  Still quite confused by what's causing that.

tests/sources/eltwise_unary_sfpu_test.cpp
```diff
@@ -63,12 +63,16 @@ void call_sfpu_operation(SfpuType operation)
             ckernel::sfpu::_init_reciprocal_<APPROX_MODE>();
             ckernel::sfpu::_calculate_reciprocal_<APPROX_MODE, 10, is_fp32_dest_acc_en>(10);
             break;
+        case SfpuType::rsqrt:
+            ckernel::sfpu::_init_sqrt_<APPROX_MODE>();
```
amahmudTT, line 67: Instead of using the same functions should we have a wrapper for rsqrt that should call the sqrt functions with the proper template arguments ? For example the initial function for both of them are the same, might be helpful if they are different. 

In other words we could have _calculate_sqrt_internal or something that has the extra boolean template argument and is called by _calculate_sqrt_ or _calculate_rsqrt_.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_recip.h
```diff
@@ -81,8 +59,9 @@ inline void _calculate_reciprocal_(const int iterations)
 template <bool APPROXIMATION_MODE>
 inline void _init_reciprocal_()
 {
-    sfpi::vConstFloatPrgm0 = 1.442695f; // ln2_recip
-    sfpi::vConstFloatPrgm1 = 2.0f;
+    sfpi::vConstIntPrgm0   = 0x7eb504f3;
```
amahmudTT, line 62: Do these numbers have special meaning ? For example in the deleted code the constant for 1/ln2 had a comment beside it. If these numbers have such meanings then corresponding comments will be helpful while going over the relevant paper (if anyone needs to go through it sometime).

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_sqrt.h
```diff
@@ -12,48 +12,60 @@ namespace ckernel
...
-            sfpi::vUInt magic = sfpi::vConstIntPrgm0;
-
-            // sqrt initial approximation
-            //  adjust bias
-            sfpi::vUInt val_s = magic + sfpi::reinterpret<sfpi::vUInt>(val);
-
-            // approximation of square root
-            val_s >>= 1;
-            sfpi::dst_reg[0] = sfpi::reinterpret<sfpi::vFloat>(val_s);
+            y = y * (sfpi::vConstFloatPrgm1 + negative_y * c);
```
amahmudTT, line 32: Curious , what is the reason of having a dedicated negative_y variable ? To make sure MAD instructions are used or SFPI does not support subtraction ?

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_sqrt.h
```diff
@@ -63,11 +75,14 @@ inline void _init_sqrt_()
 {
     if (APPROXIMATION_MODE)
     {
-        sfpi::vConstFloatPrgm0 = sfpi::s2vFloat16b(127 << 7);
+        sfpi::vConstIntPrgm0   = 0x5f0b3892;
```
amahmudTT, line 78: Same comment regarding these values having special meanings and addition of corresponding comments.

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_binary.h
```diff
@@ -167,7 +167,7 @@ inline void _sfpu_binary_init_()
 {
     if constexpr (BINOP == BinaryOp::DIV || BINOP == BinaryOp::POW)
     {
-        _init_reciprocal_<APPROXIMATION_MODE>();
+        _init_reciprocal_<false>();
```
amahmudTT, line 170: Are we intentionally ignoring the APPROXIMATE_MODE ?

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_exp.h
```diff
@@ -360,9 +360,7 @@ inline void _init_exponential_()
     }
     else
     {
-        sfpi::vConstFloatPrgm0 = 1.442695f; // ln2_recip
-        sfpi::vConstFloatPrgm1 = 2.0f;
-        sfpi::vConstFloatPrgm2 = 0.863281f;
+        _init_reciprocal_<false>();
```
amahmudTT, line 363: What are the heuristics/reasonings for choosing approximate mode vs non-approximate mode ? You could make it clear in a comment somewhere.

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_recip.h
```diff
@@ -83,8 +60,9 @@ inline void _calculate_reciprocal_(const int iterations)
 template <bool APPROXIMATION_MODE>
 inline void _init_reciprocal_()
 {
-    sfpi::vConstFloatPrgm0 = 1.442695f; // ln2_recip
-    sfpi::vConstFloatPrgm1 = 2.0f;
+    sfpi::vConstIntPrgm0   = 0x7eb504f3;
```
amahmudTT, line 63: Same comment as before.

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_sqrt.h
```diff
@@ -12,48 +13,60 @@ namespace ckernel
...
-                sfpi::dst_reg[0] = approx * val;
-            }
-            v_endif;
+        if constexpr (RECIPROCAL)
+        {
+            y = y + (sfpi::vConst1 + negative_x * y * y) * half_y;
+        }
+        else
+        {
+            half_y = -half_y;
```
amahmudTT, line 54: Why flip sign ? If it is for using fused multiply and add, did you check if it does not work when using subtraction ? By dumping the assembly ? If it works but the compiler is not smart enough then it might just be negating the number and wasting cycles.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_sqrt.h
jasondavies: I must admit, I don't particularly trust SFPI at the moment to do what I would expect w.r.t optimisation.  In the above example, _subtraction_, i.e. `sfpi::vConstFloatPrgm1 - y * c`, actually generates a whole extra SFPMUL (`-1.0 * y`).

There is also somewhat of a case for readability with optimisation in mind: the preceding operation (`c = x * y`) takes 2 cycles, which means a NOP is required unless the next instruction is non-dependent, which luckily for us, `negative_y = -y` is (which emits a 1-cycle SFPMOV with negation flag).  In other words, this is what SFPI _should_ (probably) emit here, and making it explicit is a little clearer for anyone reading it with an optimisation mindset.

Note also that the code was written primarily with Wormhole in mind, where MAD does not support negation of the multiplication or addend.  I believe this is supported in Blackhole, so there may be some room to improve marginally here, but I don't have access to confirm.

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_binary.h
jasondavies: Yes: this initialisation is specifically for the use of `_sfpu_reciprocal_` in either DIV/POW.  In both cases the default param (`APPROXIMATE=false`) is used.

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_exp.h
jasondavies: This initialisation sets constants for the use of _sfpu_reciprocal_<false>.  Happy to add a comment explaining this.

jasondavies:
> an explanation as to why we use approximate mode and why not in certain cases (where there is no definite rule) would be helpful

I'm personally not a big fan of the way this vague "approximate mode" flag (and even more confusingly, "fast approx" for exp) is used everywhere but only actually used in some places.

Specifically here, I'm calling the initialisation function for whatever function is being used in the main body.  Which seems to be `_sfpu_reciprocal_<APPROXIMATE = false>` (again, lots of confusion thanks to default template parameters etc.)

I've added comments that hopefully make this clearer.

tests/sources/eltwise_unary_sfpu_test.cpp
jasondavies: Seems reasonable.  Done!

amahmudTT:
LGTM. Thanks for addressing the comments. Will approve once tests pass, I am assuming all post commit tests and all post commit blackhole tests will be run.

jasondavies:
> I am assuming all post commit tests and all post commit blackhole tests will be run.

Thanks.  Do you mean the tt-metal integration tests?  These won't pass without tenstorrent/tt-metal#21623 - this is due to hardcoded constants being present in tt-metal.

jasondavies:
Integration tests failed as expected:

1. We have duplicate trig code that causes all integration tests with tt-metal main to fail right now (I've had to add a workaround that removes the duplicate code in some of my tt-metal PRs).
2. _This_ PR specifically requires the changes in tenstorrent/tt-metal#21623 (removes hardcoded constants in tt-metal).

fvranicTT:
@jasondavies Conflicts between metal and llk have been solved. Could you please rebase? Thank you for your patience.

fvranicTT:
@jasondavies Got it. But we also have another breaking change that just landed in the llk:
https://github.com/tenstorrent/tt-llk/pull/282. This one needs to go in a lockstep with https://github.com/tenstorrent/tt-metal/pull/22860
Let’s wait for it and then merge your changes in another lockstep.

amahmudTT (approved):

jasondavies:
LLK tests are green, this should be ready to merge.

Integration tests on this PR will fail as this needs https://github.com/tenstorrent/tt-metal/pull/21623 to pass - maybe integration tests labels should be removed.

We can run tests on the tt-metal PR once this is merged to main.

ldjurovicTT (approved):

ldjurovicTT:
@jasondavies As far as tests LLK are considered I approved. Let's just see what @fvranicTT says about integration with metal

fvranicTT:
> @jasondavies As far as tests LLK are considered I approved. Let's just see what @fvranicTT says about integration with metal

That's clear. Integration tests will fail without the metal change. I propose we merge them in a lockstep. I removed the integration tests labels and added other labels indicating what was changed. This is usually handled by the bot, but it doesn't work with PRs coming from a fork.

nvelickovicTT (approved):

jasondavies:
@fvranicTT I might take a bit more time to investigate the softmax issue, I believe it's an edge case due to handling of `±inf` inputs, resulting in unexpected results for `1/-inf`.  Should I just mark the PR as draft until resolved?

jasondavies:
Ready for review.  Latest updates:

1. Reciprocal now handles values whose absolute bit representation is larger than the magic constant (this is a small range of large floating point values whose reciprocals are very small anyway) and `±inf`, returning `0` for these cases.
2. Similarly, for sqrt/rsqrt, I've added proper handling for both `0` and `+inf`.

These changes add a small number (4-5) of operations to the original routines.  We could of course add various toggles to enable/disable these checks, but I don't think it's worth doing at the moment.

For reciprocal, at some point I could potentially look at adding support for these large floats at the edge of the supported range if necessary, but I think that's probably out of scope for this PR.

And of course, there may be some further optimisations that can be applied, but this is already (a) close to maximal precision and (b) quite optimised already.

For sqrt/rsqrt, negative inputs are not supported right now.  Adding even more checks is possible if desired (e.g. with `nan` for negative inputs) but probably best to handle in a separate PR.  Due to the way that some format conversions mangle `nan` (converting to `inf`) it's not totally straightforward.

jasondavies:
@fvranicTT Sorry, noticed an issue with the recently added `_calculate_sqrt_body_` calls in `ckernel_sfpu_trigonometry.h` and just pushed a fix.  I don't know if you need to re-approve to get it into the merge queue or not.

fvranicTT:
> @fvranicTT Sorry, noticed an issue with the recently added `_calculate_sqrt_body_` calls in `ckernel_sfpu_trigonometry.h` and just pushed a fix. I don't know if you need to re-approve to get it into the merge queue or not.

@jasondavies I am rerunning the tests now. But we are experiencing some issues with wormhole runners at the moment, so we can’t merge this one (or any other PR) yet. We are working on it.

jasondavies:
@fvranicTT Ah, thanks.  I only noticed this particular issue when integration testing with tt-metal locally.  I guess asinh/acosh aren't yet covered by tt-llk tests.

jasondavies:
Updates:

- Fixed a register pressure issue in recip (`ttnn/unit_tests/operations/eltwise/test_binaryng_fp32.py::test_pow_fp32`).
- Fixed a self-referential include problem (`ttnn/unit_tests/operations/eltwise/test_binary_bcast.py::test_binary_scalar_ops`).
- Fixed broken BH asinh (unrelated to my changes).

fvranicTT:
@jasondavies is this ready to be merged? Do you have a metal counterpart ready?

jasondavies:
@fvranicTT Yes, ready to merge. Metal PR is https://github.com/tenstorrent/tt-metal/pull/21623

jasondavies:
I took the opportunity to use the new `SFPARECIP` instruction for the BH implementation of reciprocal.

One issue I've noticed is that reciprocal should probably go back to using the `max_iter` template parameter:

- bfloat16 - this can use 0 iterations (<2 ulps) or 1 iteration (0 ulps)
- float16/tf32 - 1 iteration (0 ulps)
- float32 - 2 iterations (<1 ulps)

It's unclear what exactly the use case is for approx_mode, but presumably something like this works:

- approx_mode = true, always use 0 iterations
- approx_mode = false, fp32_dest_acc_en = false: 1 iteration
- approx_mode = false, fp32_dest_acc_en = true: 2 iterations

jasondavies:
Reciprocal is now completely rewritten for WH, which shaves off 2 cycles _and_ improves ±0, ±inf handling:

- float32 (max_iter=2): 20 cycles, correct for all normal numbers, ±0, ±inf
- bfloat16/float16 (max_iter=1): 16 cycles, edge case (same as BH) ±8.50705917302346158658e+37 → ±0 instead of ±1.1754943508222875e-38, otherwise correct for all normal numbers, ±0, ±inf

jasondavies:
@fvranicTT Ready for another CI run here, thanks.

nmauriceTT:
@jasondavies 

The SDXL failures were also happening on [main](https://github.com/tenstorrent/tt-metal/actions/runs/17308917838/job/49177456368) and don't seem related to the PR. And the Models Perf tests failed due to a missing file (also seems unrelated). 

> I do have further optimisation work in the pipeline, but I think the fundamental issue of precision has been resolved by this PR, so assuming everybody is happy with CI on the tt-metal side, it should be good to go.

Yes. That would be great.

PR #250, opened 2025-05-27
Add 32-bit support for transpose_of_faces to A2D datacopy.

jasondavies:
### Ticket

tenstorrent/tt-metal#21270

### Problem description

This is a slightly more efficient way to do transpose_wh_tile, by doing the middle face swap via unpack_A, which leaves only the within_face_transpose parts remaining.

### What's changed

- Add support for transpose_of_faces for 32-bit inputs to unpack_A (previously this was ignored).
- Add new transpose_of_faces template parameter to transpose_dest LLK API.  Important: this is currently ignored for non-32-bit data.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
@@ -25,9 +25,7 @@ inline void transpose_dest_configure_mop();
 template <bool transpose_of_faces = true, bool is_32bit = false>
 inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 {
-    math_unpack_to_dest_math_ready();
-    math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32, UnpackToDestEn>(dst_index);
-    math::math_unpack_to_dest_tile_ready();
+    math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32>(dst_index);
 
     TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU);
 
@@ -51,7 +49,8 @@ inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 
     // clear SrcA, SrcB valids; reset SrcA, SrcB, Dst counters to zero
     TTI_SETRWC(p_setrwc::CLR_AB, 0, 0, 0, 0, p_setrwc::SET_ABD);
-    //TTI_CLEARDVALID(0, 1);
+    // completely reset SrcA/SrcB sync mechanism: see https://github.com/tenstorrent/tt-metal/issues/22383
+    TTI_CLEARDVALID(0, 1);
 }
 
 template <bool is_32bit>
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
@@ -36,32 +36,17 @@ inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
         if constexpr (transpose_of_faces)
         {
             // 4x 32b face transpositions followed by 8x middle-face row swaps.
-            //ckernel_unpack_template::run(instrn_buffer, 12, 0xff0);
+            ckernel_unpack_template::run(instrn_buffer, 12, 0xff0);
         }
         else
         {
             // 4x 32b face transpositions.
-            //ckernel_unpack_template::run(instrn_buffer, 4, 0);
-          for (int i = 0; i < 4; ++i)
-          {
-              for (int dest_32b_lo = 0; dest_32b_lo < 2; ++dest_32b_lo)
-              {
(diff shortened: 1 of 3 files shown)


tt_llk_wormhole_b0/llk_lib/llk_unpack_A.h
```diff
@@ -75,10 +77,24 @@ inline void _llk_unpack_A_mop_config_(
...
+        {
+            const uint32_t outerloop = num_faces < 4 ? 1 : 2;
+            const uint32_t innerloop = num_faces < 2 ? 1 : 2;
+            ckernel_template tmp(outerloop, innerloop, num_faces > 2 ? unpack_srca_to_dest_transpose_of_faces : unpack_srca_to_dest);
+            if (num_faces > 2)
+            {
+                tmp.set_end_op(TT_OP_SETADCZW(p_setadc::UNP_A, 0, 2, 0, 1, 0b0101));
+            }
+            tmp.program(instrn_buffer);
+        }
```
jasondavies, line 90: I think I can simplify this as follows:

```cpp
if (transpose_of_faces && num_faces > 2)
{
    const uint32_t outerloop = 2;
    const uint32_t innerloop = 2;
    ckernel_template tmp(outerloop, innerloop, unpack_srca_to_dest_transpose_of_faces);
    tmp.set_end_op(TT_OP_SETADCZW(p_setadc::UNP_A, 0, 2, 0, 1, 0b0101));
    tmp.program(instrn_buffer);
}
```

I originally just copied the same approach as the non-32-bit transpose_of_faces code further down.  However, I don't see how this would work for 3 faces, presumably that's not supported at all?

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
```diff
@@ -31,8 +31,16 @@ inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 
     if constexpr (is_32bit)
     {
-        // 4x 32b face transpositions followed by 8x middle-face row swaps.
-        ckernel_unpack_template::run(instrn_buffer, 12, 0xff0);
+        if constexpr (transpose_of_faces)
```
nvelickovicTT, line 34: This is the part I am not sure I understood. This chooses between 2 implementations of the same thing, right?
First is: `unpack A to dest, transpose within face in srcB, transpose faces with SFPU`.
Second: `unpack A to dest with face transpose, transpose within face in srcB`.
If my understanding is correct I think there should only be `Second` implementation.

nvelickovicTT:
> Would be great to pick a way forward and finally resolve the int32 permute bounty issue.
> 
> Clearly adding transpose_of_faces for 32-bit to unpack is the right way forward for permute/transpose (the subject of this PR).
> 
> The following questions are unresolved:
> 
> 1. `transpose_wh_dest` (in tt-metal) and `transpose_dest` (in tt-llk) were originally added to support row-wise cumsum.  It's unclear to me if these APIs are really used for anything else.  If you do want 32-bit `transpose_dest` (to support fp32 row-wise cumsum), then you'll want to keep the approach of supporting 32-bit values for `transpose_dest` (which operates on 32x32 tiles).  The approach of this PR, is to maximise optionality, via the template parameters `<bool transpose_of_faces, bool is_32bit>` for 32x32 tiles.  This means we can reuse it in combination with `unpack(transpose_of_faces=true)`.  Just to make it absolutely clear: there are two use cases here:
> 
> * transpose_dest used for row-wise cumsum, in which case we set `transpose_of_faces=true`.
> * 32b permute/transpose (bounty issue) needs `unpack(transpose_of_faces=true)` + `transpose_16x16_face_dest` (which could be a separate API, or just supported by `transpose_dest(transpose_of_faces=false)` as per this PR)
> 
> 2. I don't really understand what `transpose_of_faces` is supposed to do for `num_faces < 4`?  So maybe it should just be supported for `num_faces == 4`.
> 
> If the goal is to minimise code complexity, we should consider dropping transpose_wh_dest/transpose_dest unless it's really important. I don't have a strong opinion and I just want to get things moving forward.

@jasondavies for point 2, it is fine to support `num_faces==4` for the time being.
As for point 1, I can't find what the question is?

nvelickovicTT:
Actually, let me ask you like this. If you remove completely the option to transpose faces using SFPU, and just leave the option to do it using `unpack(transpose_of_faces=true)`, will that break anything currently?

jasondavies:
> As for point 1, I can't find what the question is?

Are you happy with the approach I've taken, which is to add a template param `<bool transpose_of_faces>` to `_llk_transpose_dest_`?

This preserves the ability to do a full 32x32 transpose in dest.

I guess maybe in the interest of progress, we should just go ahead with this and if we think the API surface is too broad we can remove later.

nvelickovicTT:
I've put it on auto-merge. Once you resolve comment, it will merge as is. But before that, please add the comment with a link to issue.

jasondavies:
> I've put it on auto-merge. Once you resolve comment, it will merge as is. But before that, please add the comment with a link to issue.

Thanks, I think you need to approve again as I don't have permissions.  Just pushed a commit linking to the API design question.  Once this is merged I'll update my other PR (copying implementation to Blackhole) with the same comment.

nvelickovicTT (approved):

PR #274, opened 2025-05-29
Foreshadow load_replay_buf api change

nathan-TT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/22101

### Problem description
In providing replay intrinsics, it makes sense to simplify LLK's replay API.  There happens to be a use of this API directly from tt_metal, so we need to do a bit of preparatory work to get the new API in place.  (The alternative would be to do some cleanup after the fact, ETA: but IMHO this ordering is cleaner).

### What's changed
Add a new `load_replay_buf` overload that only templatize the exec-while-loading parameter. Once tt_metal updates its tt_llk submodule, we can then change it to use this new overload. That will then allow us to update tt_llk to use the new replay intrinsics.

### Type of change

- [] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### Checklist

- [YES] [All post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml) CI passes
- [YES] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml) CI passes (if applicable)

### DIFF
--- a/tt_llk_blackhole/common/inc/ckernel.h
+++ b/tt_llk_blackhole/common/inc/ckernel.h
@@ -683,32 +683,42 @@ inline void enable_gathering()
             : "t1");
 }
 
-// Pass a lambda function (or a regular function pointer) that takes void,
-// returns void, and issues the instructions you want to load into the
-// replay buffer. start, len, and exec_while_loading have the same meaning
-// as they do for the REPLAY instruction, as descired in assembly.yaml.
-template <uint start, uint len, bool exec_while_loading = false, typename F>
-inline void load_replay_buf(F fn)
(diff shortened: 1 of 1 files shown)


nvelickovicTT:
@nathan-TT if I didn't miss anything, the only instance of `load_replay_buf` is in `llk_unpack_tilize_api.h`, and this is going to be cleaned up with https://github.com/tenstorrent/tt-metal/pull/22638.
So, to simplify your changes, you could wait for the 22638 to go in.

nathan-TT:
No longer needed, tt_metal no longer uses the function at issue.

PR #280, opened 2025-06-01
Fix pack_untilize for block_ct_dim != full_ct_dim on BH

pmilenkovicTT:
### Ticket
[#11573](https://github.com/tenstorrent/tt-metal/issues/11573)

### Problem description
Untilize operation can be performed either using the slow algorithm (unpack_untilize) or by using the fast algorithm (pack_untilize). However, pack_untilize had the limitation of input tensor fitting only in one Destination register bank (8 tiles width). The goal of this fix is to enable the pack_untilize operation to work for any input size, by fixing the block_ct_dim != full_ct_dim code path, where block_ct_dim is the number of tiles in the single Destination register bank, and full_ct_dim is the width of the whole input expressed in tiles.

### What's changed

- Rewrote the algorithm to execute everything in the MOP - no more handling of the last row in main
- Removed unnecessary address modes
- Added static asserts for the unsupported inputs

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_pack_untilize.h
+++ b/tt_llk_blackhole/llk_lib/llk_pack_untilize.h
@@ -89,7 +89,6 @@ inline void _llk_pack_untilize_mop_config_(
     load_replay_buf(
         ckernel::packer::replay_buf_offset,
         replay_buf_len,
-        false,
         []
         {
             // Update L1 address
@@ -100,7 +99,7 @@ inline void _llk_pack_untilize_mop_config_(
         });
 
     // After the inner loop finishes, move to the next row in the block, and update L1 address.
-    tmp.set_end_ops(TT_OP_INCADCXY(p_setadc::PAC, 0, 0, 1, 0), TT_OP_REPLAY(ckernel::packer::replay_buf_offset, replay_buf_len, 0, 0));
+    tmp.set_end_ops(TT_OP_INCADCXY(p_setadc::PAC, 0, 0, 1, 0), lltt::replay_insn(ckernel::packer::replay_buf_offset, replay_buf_len));
 
     /*
     Close the row in the block by setting the Last bit to 1 in the last inner loop instruction.


tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -144,12 +144,19 @@ inline void _llk_pack_untilize_init_(
     uint y_stride       = FACE_C_DIM * x_stride;
     const uint z_stride = 2 * face_r_dim * y_stride;
     cfg_reg_rmw_tensix<PCK0_ADDR_CTRL_ZW_REG_0_Zstride_RMW>(z_stride);
+    std::uint32_t output_addr_offset;
 
-    if (block_ct_dim != full_ct_dim)
+    if constexpr (narrow_row)
```
fvranicTT, line 149: May I suggest IILE (immediately invoked lambda expression)?
```cpp
const std::uint32_t output_addr_offset = []() {
    if constexpr (narrow_row) {
        return SCALE_DATUM_SIZE(pack_dst_format, full_ct_dim * row_num_datums);
    }
    
    // Pre-compute the face multiplier at compile time, if possible
    constexpr uint32_t face_multiplier = (num_faces > 1) ? (num_faces >> 1) : 1;
    return SCALE_DATUM_SIZE(pack_dst_format, full_ct_dim * face_multiplier * FACE_C_DIM);
}();
```

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -135,6 +133,8 @@ inline void _llk_pack_untilize_init_(
     const std::uint32_t pack_src_format, const std::uint32_t pack_dst_format, const std::uint32_t face_r_dim = FACE_R_DIM, const std::uint32_t num_faces = 4)
 {
     static_assert(!diagonal, "Diagonal not supported");
+    static_assert(block_ct_dim <= 8, "block_ct_dim must be less than or equal to 8");
```
fvranicTT, line 136: Let's group all checks together. You can add:
```cpp
if constexpr (narrow_row) {
    static_assert(row_num_datums < FACE_C_DIM, 
        "row_num_datums must be less than FACE_C_DIM for narrow_row packing");
}
```
As a bonus, you can use FACE_C_DIM instead of relying on magic numbers 😉

pmilenkovicTT: Done!

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
pmilenkovicTT: Tried it with @fvranicTT, didn't make sense as it doesn't increase perf.

fvranicTT: The whole idea was having constexpr lambda, but it's not possible :(

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -43,86 +34,93 @@ inline void _llk_pack_untilize_mop_config_(
...
+            TTI_WRCFG(p_gpr_pack::OUTPUT_ADDR, 0, THCON_SEC0_REG1_L1_Dest_addr_ADDR32);
+            TTI_NOP;
+        });
+
+    tmp.set_end_ops(
+        TT_OP_INCADCXY(p_setadc::PAC, 0, 0, 1, 0),                             // inc ch0_y counters
+        TT_OP_REPLAY(ckernel::packer::replay_buf_offset, replay_buf_len, 0, 0) // update row address
+    );
+
+    tmp.set_last_inner_loop_instr(TT_OP_PACR(
```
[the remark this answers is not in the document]
pmilenkovicTT: Done.

rtawfik01 (approved):

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -25,104 +25,111 @@ inline void _llk_pack_untilize_configure_addrmod_()
...
+            0,
+            PACK_INTF_SEL,
+            0,
+            0,
+            p_pacr::NO_CTXT_CTRL,
+            0,
+            0));
+
+    /*
+    Since there are two inner loop operations, the instruction set by set_last_inner_loop_instr
```
amahmudTT, line 73: I think you are updating the L1 address in the end op and not in the last instruction. The comments are okay, but you could put them in the correct place.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -25,104 +25,111 @@ inline void _llk_pack_untilize_configure_addrmod_()
...
+    will replace the second inner loop operation (in the last iteration, call the PACR instuction
+    with the Last bit set to 1 instead of 0 to close the row and allow L1 address to be updated).
+    Therefore, by setting the W counter to maxium value (15) as a start operation,
+    TT_OP_INCADCZW in the inner loop will reset it to 0 in the first iteration of the inner loop.
+    */
+    tmp.set_start_op(TT_OP_SETADC(p_setadc::PAC, p_setadc::CH_0, p_setadc::SET_W, 15));
+
+    const std::uint32_t replay_buf_len = 4;
```
amahmudTT, line 81: You can put the part of the comment above here, and refer to this as the end op not the last instruction. AFAIK end op is the one that will execute after the inner loop is finished, and in that one you are updating the address.  But in the last instruction ( which replaces the last executed instruction at the termination of the loop , you are using the packer with the last bit set to 1). So if I am not wrong, we need to 

Refer to end op instead of last instruction when needed
Put the comments at the right places.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -25,104 +25,111 @@ inline void _llk_pack_untilize_configure_addrmod_()
...
+        0,
+        PACK_INTF_SEL,
+        0,
+        0,
+        p_pacr::NO_CTXT_CTRL,
+        0,
+        1));
+
+    tmp.set_last_outer_loop_instr(TT_OP_PACR(
+        p_pacr::CFG_CTXT_0,
```
amahmudTT, line 119: I think you have the last inner loop instruction and the last outer loop instruction the same, in that case you could save it in a constexpr uint32_t and then assign them using the same uint. Reducing code size.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -135,6 +142,12 @@ inline void _llk_pack_untilize_init_(
     const std::uint32_t pack_src_format, const std::uint32_t pack_dst_format, const std::uint32_t face_r_dim = FACE_R_DIM, const std::uint32_t num_faces = 4)
 {
     static_assert(!diagonal, "Diagonal not supported");
+    static_assert(block_ct_dim <= 8, "block_ct_dim must be less than or equal to 8");
```
amahmudTT, line 145: You could add ", on blackhole due to dest register size limit" -> if that is the case.

amahmudTT:
> Hey @amahmudTT I added some more comments to explain the algorithm, would appreciate if you could take another look and see if it makes more sense now.

I think it is much more understandable now. But still left some comments.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
pmilenkovicTT: Reworded the sentence about the L1 address, the point was to explain why the value of the W counter is 15 after the reset.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
pmilenkovicTT: I'm not sure it is possible - the instruction depends on PACK_INTF_SEL, which depends on narrow_row and num_faces, which are not compile time. 
Once we actually start reading stuff from CBs, it will be possible to do it, so I can leave a comment to revisit it maybe?

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -25,104 +25,111 @@ inline void _llk_pack_untilize_configure_addrmod_()
...
-        .z_src = {.clr = 1},
-    }
-        .set(ADDR_MOD_2);
 }
 
 template <std::uint32_t block_ct_dim, std::uint32_t full_ct_dim = block_ct_dim, bool diagonal = false>
 inline void _llk_pack_untilize_mop_config_(
     const std::uint32_t face_r_dim = FACE_R_DIM, const std::uint32_t num_faces = 4, bool narrow_row = false, std::uint32_t row_num_datums = TILE_C_DIM)
 {
+    /*
```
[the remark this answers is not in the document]
pmilenkovicTT: It is mentioned in the _llk_pack_untilize_, but I added it on top of the mop config as well.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
pmilenkovicTT: Done

PR #289, opened 2025-06-03
initial debug functions for halting tensix ( from tt-metal )

ldjurovicTT:
### Ticket
None

### Problem description
We want to be able to halt single and all threads on Tensix to see what is src and dst registers while debugging

### What's changed
Added header file that extracts some tt-metal functions and adapted it to be LLK standalone

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/helpers/include/dbg_halt.h
+++ b/tests/helpers/include/dbg_halt.h
@@ -1,0 +1,53 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+template <ckernel::ThreadId thread_id>
+inline void dbg_thread_halt()
+{
+    static_assert(
+        (thread_id == ckernel::ThreadId::MathThreadId) || (thread_id == ckernel::ThreadId::UnpackThreadId) || (thread_id == ckernel::ThreadId::PackThreadId),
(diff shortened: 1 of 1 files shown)


tests/helpers/include/dbg_halt.h
```diff
@@ -0,0 +1,53 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+template <ckernel::ThreadId thread_id>
+inline void dbg_thread_halt()
+{
+    static_assert(
+        (thread_id == ckernel::ThreadId::MathThreadId) || (thread_id == ckernel::ThreadId::UnpackThreadId) || (thread_id == ckernel::ThreadId::PackThreadId),
+        "Invalid thread id set in dbg_wait_for_thread_idle(...)");
+
+    if constexpr (thread_id == ckernel::ThreadId::UnpackThreadId)
```
fvranicTT, line 11: What about PatckThreadId?

ldjurovicTT: This is how it is implemented in metal...

fvranicTT: So `dbg_halt_pack()` effectively does nothing 😁

fvranicTT: In that case, could you remove dbg_halt_pack() since it doesn’t do anything. Once you have it implemented we could add it back. The way it is now, it will only confuse people.

Issue #300, opened 2025-06-04
[ Test Infra ] Skip fidelity sweep for ops not affected by fidelity


ldjurovicTT:
We are running a lot of tests inside of a sweep that change fidelity on ops like eltwise add which is not affected by fidelity. Those should be skipped, and will also lower needed time for running them

PR #304, opened 2025-06-05
Skipping tests that are not elwmul and sweep fidelity

ldjurovicTT:
### Ticket
#300 

### Problem description
Some tests are running fidelity sweep for operations not affected by math fidelity

### What's changed
Skipped those tests

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/test_multiple_tiles_eltwise.py
+++ b/tests/python_tests/test_multiple_tiles_eltwise.py
@@ -109,6 +109,10 @@ param_ids = generate_param_ids(all_params)
     ids=param_ids,
 )
 def test_multiple_tiles(testname, formats, dest_acc, mathop, math_fidelity, tile_cnt):
+
+    if mathop != MathOperation.Elwmul and math_fidelity != MathFidelity.LoFi:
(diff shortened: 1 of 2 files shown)


tests/python_tests/test_multiple_tiles_eltwise.py
```diff
@@ -109,6 +109,10 @@ def generate_golden(op, operand1, operand2, data_format, math_fidelity):
     ids=param_ids,
 )
 def test_multiple_tiles(testname, formats, dest_acc, mathop, math_fidelity, tile_cnt):
+
+    if mathop != MathOperation.Elwmul and math_fidelity != MathFidelity.LoFi:
```
fvranicTT, line 113: Let's merge this as is and try to find a way not to have these tests at all. I'd like to keep the number of skipped tests to a minimum, and only skip what is either not supported, not implemented or failing. Having 100 tests skipped just because they don't make much sense is a sign we're doing something wrong.

fvranicTT: I was trying with this approach from the start, but we had issues picking combinations. Let's park this discussion here and talk offline.

PR #359, opened 2025-06-12
Fix transpose_dest on BH.

### LINKED ISSUE #367: Diagnose transpose_dest sync issue on BH
In #359 it was found that adding a `p_stall::MATH` wait condition to the STALLWAIT for transpose_dest fixed incorrect transpose of the first face on BH.

Inserting 5x NOPs also seemed to fix the issue.

It doesn't make sense to stall MATH on MATH wait condition, and most likely this just creates a short delay.

The exact cause of the original issue should be diagnosed and a more appropriate fix used.

jasondavies:
### Problem description

This fixes two issues with `_llk_math_transpose_dest_` on Blackhole:

1. For some reason counters were not starting at zero prior to tranpose_dest, even though there is a SETRWC in _init.
2. The first face was incorrect, and this was diagnosed as some kind of race condition issue.

### What's changed

1.  Move the SETRWC call to the beginning of transpose_dest instead.  The same is done for WH even though it doesn't seem to suffer from this problem.
2. Modify the STALLWAIT to wait for SrcA/SrcB DVALID before executing MOVD2A/MOVB2A, since these instructions do not automatically wait.  Same for WH even though it doesn't seem to exhibit a problem.  Closes #367.
3. Removed the CLEARDVALID workaround on BH for https://github.com/tenstorrent/tt-metal/issues/22383 as it doesn't seem to be necessary.

### Type of change

- [ ] Documentation update

### Checklist


jasondavies:
@nvelickovicTT let me know what you think.

Once this is merged to main, we can run BH tests on https://github.com/tenstorrent/tt-metal/pull/23415 - I've verified all permute tests pass locally.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
```diff
@@ -35,8 +35,9 @@ template <bool transpose_of_faces = true, bool is_32bit = false>
 inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 {
     math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32>(dst_index);
+    math::reset_counters(p_setrwc::SET_ABD_F);
 
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH);
```
amahmudTT, line 40: I do not understand this change , are you stallling math exu till math exu is finished ? What happens when you do not put this change in ? I think the solution should be a bit different.

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
```diff
@@ -35,8 +35,9 @@ template <bool transpose_of_faces = true, bool is_32bit = false>
 inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 {
     math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32>(dst_index);
+    math::reset_counters(p_setrwc::SET_ABD_F);
 
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH);
```
amahmudTT, line 40: Same comment as above.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
jasondavies: It breaks without this change (specifically it fails to correctly transpose the first of four faces).

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
jasondavies: Strictly speaking, the change isn't necessary for Wormhole as tests already pass without changes, though I thought it best to make the same change and move SETRWC (math::reset_counters) to transpose_dest to avoid confusion, hence I might as well make the same STALLWAIT change too.

I'm personally not 100% sure why the STALLWAIT change works for BH, but it does.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
amahmudTT: If you are waiting for the reset counters to finish before using MATH, i think you need to wait for CFG to finish 
as the function uses SETC16 instructoin in one branch, but in the other it uses a mailbox, which I do not know at the moment what to do about, probably nothing.

tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h
jasondavies: > Could you try STALL_CFG instead of STALL_MATH ?

This doesn't make sense.  We definitely want to stall FPU (MATH).  I think the issue is the wait condition i.e. we want to wait for Configuration Unit to complete work.  See my other comment.

jasondavies: Yes, I think we're on the same page.  The distinction between "what are we stalling" and "why are we stalling"/"what are we waiting for" is not very clear in `p_stall`.

jasondavies:
I've also added a separate `_llk_unpack_set_srcab_dummy_valid_` to resolve the use of CLEARDVALID.  Let me know if you'd rather I did that in a separate PR.

tt_llk_blackhole/llk_lib/llk_unpack_common.h
```diff
@@ -155,9 +155,14 @@ inline void _llk_enable_int8_fpu_math_()
...
 
-inline void _llk_unpack_set_srcb_dummy_valid_()
+inline void _llk_unpack_set_srcab_dummy_valid_()
 {
     TTI_STALLWAIT(p_stall::STALL_UNPACK, p_stall::UNPACK);
     TTI_UNPACR_NOP(SrcB, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
     TTI_UNPACR_NOP(SrcA, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
 }
+
+inline void _llk_unpack_set_srcb_dummy_valid_()
```
nvelickovicTT, line 165: Here you still need the `TTI_STALLWAIT()` because `UNPACR_NOP` is issued in parallel to `UNPACR` instruction, so it might execute before unpacker finishes.

tt_llk_wormhole_b0/llk_lib/llk_unpack_common.h
```diff
@@ -156,9 +156,14 @@ inline void _llk_enable_int8_fpu_math_()
...
 
-inline void _llk_unpack_set_srcb_dummy_valid_()
+inline void _llk_unpack_set_srcab_dummy_valid_()
 {
     TTI_STALLWAIT(p_stall::STALL_UNPACK, p_stall::UNPACK);
     TTI_UNPACR_NOP(SrcB, p_unpacr_nop::UNP_SET_DVALID);
     TTI_UNPACR_NOP(SrcA, p_unpacr_nop::UNP_SET_DVALID);
 }
+
+inline void _llk_unpack_set_srcb_dummy_valid_()
```
nvelickovicTT, line 166: Same comment for `TTI_STALLWAIT()` here.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
nvelickovicTT: To unblock a whole chain of issues waiting on int32, I'm OK to use this workaround for the time being. and we will put this investigation as a higher priority issue internally.

nvelickovicTT (approved):

jasondavies:
@nvelickovicTT I'm still not 100% happy with how we set AB DVALID (is_32bit=false) or B DVALID (is_32bit=true) here.  I think this causes issues if transpose_wh_dest is used for 32-bit values.

Maybe it's too ambitious to include in this PR and we should keep CLEARDVALID here, then make a separate change for the AB/B DVALID changes?

(I'm starting to feel that maybe the cleanest API choice would be to make unary datacopy support 32-bit values, and make the transpose_dest API specifically limited to data that's already in dest.  But maybe best to do that separately.)

jasondavies:
> Maybe it's too ambitious to include in this PR and we should keep CLEARDVALID here, then make a separate change for the AB/B DVALID changes?

I've reinstated CLEARDVALID for now.

fvranicTT:
Just to confirm, are hangs in metal tests expected without changes from your branch on metal?

jasondavies:
> Just to confirm, are hangs in metal tests expected without changes from your branch on metal?

No, I think these may be related to the DVALID stuff, which I've now removed so we can deal with it separately.

lpremovicTT:
> @nvelickovicTT I'm still not 100% happy with how we set AB DVALID (is_32bit=false) or B DVALID (is_32bit=true) here. I think this causes issues if transpose_wh_dest is used for 32-bit values.
> 
> Maybe it's too ambitious to include in this PR and we should keep CLEARDVALID here, then make a separate change for the AB/B DVALID changes?
> 
> (I'm starting to feel that maybe the cleanest API choice would be to make unary datacopy support 32-bit values, and make the transpose_dest API specifically limited to data that's already in dest. But maybe best to do that separately.)

transpose_dest API was always intended to be limited to that's already in dest, but that wouldn't fix the DVALID issue we are chasing here as transpose_dest has to use src registers to perform transpose and as such has to include dummy DVALIDS to avoid clobbering data in src registers.

Either way, its fine for it to wait for another PR.

jasondavies:
I've moved CLEARDVALID to _llk_math_eltwise_unary_datacopy_init_: even though we still don't have an explanation for why a single SrcA/SrcB flip seems to break it, it seems likely that it makes some kind of implicit assumption, and this seems to be the right place to put a full SrcA/SrcB reset anyway.  I would still like to know the actual reason!

I think I've also figured out the correct wait condition for BH: since we initially use MOVD2B, this doesn't automatically wait for `SrcB[MatrixUnit.SrcBBank].AllowedClient == SrcClient::MatrixUnit` and so could lead to a race.  The same applies to WH, though it doesn't seem to have ever caused any problems.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
```diff
@@ -35,8 +35,11 @@ template <bool transpose_of_faces = true, bool is_32bit = false>
 inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 {
     math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32>(dst_index);
+    math::reset_counters(p_setrwc::SET_ABD_F);
 
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU);
+    // Wait condition SRCB_VLD is required as MOVD2B doesn't automatically wait
+    // for SrcB[MatrixUnit.SrcBBank].AllowedClient == SrcClient::MatrixUnit.
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
```
lpremovicTT, line 42: Since 16 bit path uses both SrcA and B and MOVB2A doesn't check SrcA data ready, should we also stall on SRCB_VLD? Might not be strictly necessary but i think it would help clarity.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
```diff
@@ -35,8 +35,11 @@ template <bool transpose_of_faces = true, bool is_32bit = false>
 inline void _llk_math_transpose_dest_(const std::uint32_t dst_index)
 {
     math::set_dst_write_addr<DstTileLayout::Default, DstTileShape::Tile32x32>(dst_index);
+    math::reset_counters(p_setrwc::SET_ABD_F);
```
lpremovicTT, line 38: Can this reset be moved back into init now that we have found that not stalling on SRCB_VLD was the issue?

jasondavies: No, this is a separate issue.  There were two issues originally. :)  It breaks without the counter reset here.

tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
lpremovicTT: From looking at the src_mask in assembly.yaml it seems to not wait.

jasondavies: Will add it in that case, thanks.

tt_llk_blackhole/llk_lib/llk_math_eltwise_unary_datacopy.h
```diff
@@ -265,5 +265,8 @@ inline void _llk_math_eltwise_unary_datacopy_init_(
 
     TTI_SETC16(CLR_DVALID_SrcA_Disable_ADDR32, 0);
 
+    // Unclear exactly why this is needed, see: https://github.com/tenstorrent/tt-metal/issues/22383
+    TTI_CLEARDVALID(0, 1);
```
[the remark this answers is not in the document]
jasondavies: Unfortunately it looks like this change causes timeouts in test_eltwise_unary_sfpu so will revert.

fvranicTT:
Running metal integration tests:

BH - https://github.com/tenstorrent/tt-metal/actions/runs/15682510787
WH - https://github.com/tenstorrent/tt-metal/actions/runs/15682530027

fvranicTT:
> @fvranicTT Did you mean to test against [tenstorrent/tt-metal#23415](https://github.com/tenstorrent/tt-metal/pull/23415) ? I only see tt-llk refs in your branch, not my tt-metal changes.

Ah you are right! I forgot about the branch.

jasondavies:
@fvranicTT Looks like we caught an issue in my tt-metal PR: I was attempting to disable flush denormals for all types, but this caused a unit test failure for moreh_matmul.  I've now reverted that change.  Otherwise it seems to look good barring any other ttnn unit test failures we might see on re-run with WH.

The other few failures look more like timeout/unrelated issues to me.

BH is all passing but didn't run ttnn unit tests.

PR #360, opened 2025-06-12
Replace TTI_REPLAY with lltt::record, lltt::replay

nathan-TT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/22101

### Problem description
The compiler has no visibility of TT_REPLAY use, and that interferes with its TT_REPLAY optimization pass. Ideally we'd like the compiler to generate TT_REPLAYs that do not interfere with user use.  Right now we have to disable that pass for blackhole, and it seems to get lucky on wormhole.

### What's changed
* `#include <lltt.h>` as necessary, a new low-level tt instruction header provided by sfpi
* Replace `TTI_REPLAY(start, len, exec, true)` with `lltt::record<exec>(start, len)` (in all use cases `exec` is a manifest constant)
* Replace `TTI_REPLAY(start, len, false, false)` with `lltt::replay(start, len)`
* Replace `TT_OP_REPLAY(start, len, false, false)` with lltt::replay_insn(start, len)`
* Change `load_replay_buf` function into: 
```
template <bool Exec = false, typename Callable, typename... Args>
[[gnu::always_inline, gnu::flatten]] inline void load_replay_buf(uint start, uint len, Callable&& callable, Args&&... args);
```
and similarly adjust callers (essentially the runtime `exec_while_loading` argument becomes a template parameter.

This change requires sfpi to be updated to v6.12.0, which tt_metal is now using -- not sure how tt_llk gets its sfpi release updated.

The sfpi compiler optimization no longer inserts TT_REPLAY when the function already contains user calls to the new record & replay functions. (This optimization will be extended)

### Type of change

- [YES] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -40,7 +40,7 @@ OPTIONS_ALL	:= -O3 -mabi=ilp32 -std=$(CXX_VERSION) -ffast-math $(ARCH_CPU)
 OPTIONS_COMPILE := -fno-use-cxa-atexit -Wall -fpermissive -fno-exceptions -fno-rtti -Werror \
 				   -Wno-unknown-pragmas -Wno-error=multistatement-macros -Wno-error=parentheses \
 				   -Wno-error=unused-but-set-variable -Wno-unused-variable -DTENSIX_FIRMWARE \
-				   $(ARCH_DEFINE)
+				   $(ARCH_DEFINE) -DCOMPILE_FOR_TRISC=
 OPTIONS_LINK	:= -fexceptions -Wl,-z,max-page-size=16 -Wl,-z,common-page-size=16 -nostartfiles -Wl,--trace
 
 INCLUDES := -I../$(ARCH_LLK_ROOT)/llk_lib -I../$(ARCH_LLK_ROOT)/common/inc \


tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -10,6 +10,13 @@
 // MT: This should be dissolved and moved to the appropriate place
 #include "tensix.h"
 
+// This header is include on non-trisc builds, for reasons
```
amahmudTT, line 13: nit include -> included ?

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -667,34 +674,14 @@ inline void enable_gathering()
...
-
-// Same as above, but used if start/len/exec_while_loading are not known
-// at compiletime.
-template <typename F>
-inline void load_replay_buf(uint start, uint len, bool exec_while_loading, F fn)
+#if defined(COMPILE_FOR_TRISC)
+// Place instructions into the replay buffer. EXEC is true to execute
+// when loading (default is false). START is where to place in the
+// replay buffer, and LEN is the number of instructions to record
+// (should match the expansion of FN). FN is a callable, to which ARGS
```
amahmudTT, line 681: I do not see anything named FN here as referred by your comment above.

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -667,34 +674,14 @@ inline void enable_gathering()
...
+// are forwarded.
+template <bool Exec = false, typename Callable, typename... Args>
```
amahmudTT, line 683: There has been some talk regarding not using bool as arguments as it gets confusing and to use enum instead for better readability. Do you want to do that from the get go in this case ?

nvelickovicTT (approved):

nathan-TT:
The tests are failing because the test harness is not defining `COMPILE_FOR_TRISC`, which we check in `ckernel.h` before including `lltt.h`.  I found I needed that check because `ckernel.h` was being included on non-trick compiles.
While strictly speaking we should define it to 0, 1, or 2 for the unpack, math and pack cores respectively, but I think just defining it to nothing will be sufficient (and break many uses that expect it to have a value).

PR #364, opened 2025-06-13
[Test Infra] Refactor format handling to reduce cross-file dependency, duplicate logic & tight coupling

skotaracTT:
### Ticket

https://github.com/orgs/tenstorrent/projects/123/views/1?pane=issue&itemId=114761584&issue=tenstorrent%7Ctt-llk%7C353
### Problem description

As support for more data formats grows, we've observed increasing code bloat, duplication, and tight coupling—particularly in areas like `params.h`. Adding or updating a format often requires touching multiple layers and files, leading to code bloat, repetition, friction and maintenance overhead.
This refactor restructures the format-handling logic to be more **flexible** and **extensible**:
- Reduced boilerplate and repetitive logic
- Minimize cross-file dependencies
- Make it easier to scale as new formats are introduced
- Improve overall readability and structure

This sets the foundation for cleaner, faster development as the system evolves.

### What's changed

This refactor removes ~330 lines of redundant and tightly coupled code, resulting in a cleaner and more modular format handling structure.
### Type of change

- [ ] Documentation update


tests/python_tests/helpers/test_config.py
```diff
@@ -89,18 +83,29 @@ def generate_build_header(
     header_content.extend(["", "// Data format configuration"])
     formats = test_config.get("formats")
     if isinstance(formats, InputOutputFormat):
-        header_content.append(f"#define {unpack_A_src_dict[formats.input_format]}")
-        header_content.append(f"#define {pack_dst_dict[formats.output_format]}")
+        header_content.extend(
```
fvranicTT, line 86: You could group this into one extend:
```python
header_content.extend([
    "// Activating Data Format Inference Model\n",
    "#define DATA_FORMAT_INFERENCE_MODEL true",
    f"constexpr auto UNPACK_A_IN = static_cast<std::underlying_type_t<DataFormat>>(DataFormat::{formats.input_format.name});",
    f"constexpr auto PACK_OUT = static_cast<std::underlying_type_t<DataFormat>>(DataFormat::{formats.output_format.name});"
])
```

tests/helpers/include/params.h
```diff
@@ -40,254 +40,7 @@ constexpr bool dest_acc_en_input =
...
-#endif
-#if defined(MATH_UINT8)
-MATH_CASE(UInt8)
-#endif
-
-#undef MATH_CASE
-
-#if !defined(MATH_BFP8_B) && !defined(MATH_INT32) && !defined(MATH_FLOAT32) && !defined(MATH_FLOAT16) && !defined(MATH_FLOAT16_B) && !defined(MATH_TF32) && \
-    !defined(MATH_UINT32) && !defined(MATH_UINT16) && !defined(MATH_INT8) && !defined(MATH_UINT8)
+#if DATA_FORMAT_INFERENCE_MODEL
```
fvranicTT, line 43: If I'm reading this well, you don't need lines 47-51 anymore, right? Because these will be defined in the build.h.

tests/python_tests/helpers/test_config.py
skotaracTT: Agreed 🫡

tests/helpers/include/params.h
skotaracTT: Not quite... Logically yes, however there is a slight trick. Since formats need to be passed into llk functions as constants they are defined once. In our testing infra there are two cases when we define llk pipeline formats
1. When we hardcode an exact combination we want, in which case we have defined all the formats in our python test file before we run the test. These 7 formats are directly defined in `build.h`.
2. We define an input and output format, in which case the rest of the pipeline formats need to be inferred with the data format inference model. In this case only `UNPACK_A_IN` and `PACK_OUT` are defined in `build.h`. The rest are awaiting decision from `get_data_formats`.

tests/helpers/include/params.h
```diff
@@ -40,254 +40,7 @@ constexpr bool dest_acc_en_input =
 
 constexpr bool unpack_to_dest = UNPACKING_TO_DEST;
 
-#define UNPACK_A_SRC_CASE(data_format) constexpr auto UNPACK_A_IN = get_data_format(DataFormat::data_format);
```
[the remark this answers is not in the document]
skotaracTT: No greater satisfaction!🤩😮‍💨😎

tests/helpers/include/params.h
fvranicTT: Ah, got it. Thanks!

PR #373, opened 2025-06-17
test: Binary int32 testing

Aswinmcw:
### Ticket
N/A

### Problem description
binary ops testing added only for float

### What's changed
Adds int dataformat for testing with some improvements

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/test_sfpu_binary.py
+++ b/tests/python_tests/test_sfpu_binary.py
@@ -25,7 +25,8 @@ from helpers.test_config import generate_make_command
 from helpers.utils import passed_test, run_shell_command
 
 # SUPPORTED FORMATS FOR TEST
-supported_formats = [DataFormat.Float16_b, DataFormat.Int32]  # , DataFormat.Float16]
+supported_float_formats = [DataFormat.Float16_b]  # , DataFormat.Float16]
+supported_int_formats = [DataFormat.Int32]
 
 #   INPUT-OUTPUT FORMAT SWEEP
 #   input_output_formats(supported_formats)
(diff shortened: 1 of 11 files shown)


tests/python_tests/test_sfpu_binary.py
```diff
@@ -69,18 +76,34 @@ def generate_golden(operation, operand1, operand2, data_format):
...
-    ],
+    input_output_formats(supported_float_formats),
+    dest_acc=[DestAccumulation.No],
+    mathop=float_ops,
 )
+
+int_params = generate_params(
+    ["sfpu_binary_test"],
+    input_output_formats(supported_int_formats),
+    dest_acc=[DestAccumulation.No],
```
ldjurovicTT, line 101: If we are using Int32 DestAccumulation should be set to Yes. It passes for now because range of generated int numbers doesn't exploit it's full range. But if we leave it like this it may cause some problems later

ldjurovicTT (approved):
From the side of tests I can approve. We should wait for codeowners for other folders to approve also

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_shift.h
```diff
@@ -48,8 +48,14 @@ inline void _calculate_binary_right_shift_(const uint dst_offset)
         TTI_SFPLOAD(0, 12, ADDR_MOD_7, 0);
         TT_SFPLOAD(1, 12, ADDR_MOD_7, dst_offset * dst_tile_size);
         TTI_SFPMOV(0, 0, 4, 0); // save shift_value for later
+        // if (shift_amount < 0 OR shift_amount >= 32) -> result should be 0
+        TTI_SFPSETCC(0, p_sfpu::LREG1, p_sfpu::LREG0, 4);
+        TTI_SFPIADD(0xFE0, p_sfpu::LREG1, p_sfpu::LREG2, 1); // 0xFE0 = -32
+        TTI_SFPCOMPC(0, p_sfpu::LREG0, p_sfpu::LREG0, 0);
```
rdjogoTT, line 54: this combination of: SFPIADD+SFPCOMPC can be achieved by setting the instr_mod of SFPIADD to 9 instead of 1. "condition code register is set to complement of MSB of addition" for 9 rather than "condition code register is set to MSB of addition result" for 1.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_shift.h
```diff
@@ -48,8 +48,14 @@ inline void _calculate_binary_right_shift_(const uint dst_offset)
...
+        TTI_SFPMOV(0, 9, p_sfpu::LREG0, 0);
```
rdjogoTT, line 55: `p_sfpu::LCONST_0` == 9, so you can put it here. Helps with understanding as well

PR #376, opened 2025-06-17
Add default LLK_HOME

sstanisicTT:
### Problem description
Sets default LLK_HOME in case it's not provided by environment

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/conftest.py
+++ b/tests/python_tests/conftest.py
@@ -6,6 +6,7 @@ import os
 import subprocess
 import sys
 import time
+from pathlib import Path
 
 import pytest
 import requests
@@ -20,10 +21,7 @@ from helpers.perf import delete_reports
 def init_llk_home():
     if "LLK_HOME" in os.environ:
         return
-
-    os.environ["LLK_HOME"] = os.path.abspath(
-        os.path.join(os.path.dirname(__file__), "..", "..", "..")
-    )
+    os.environ["LLK_HOME"] = str(Path(__file__).resolve().parents[2])
 
 
 def set_chip_architecture():


tests/python_tests/conftest.py
```diff
@@ -17,6 +17,20 @@
 from helpers.perf import delete_reports
 
 
+def init_llk_home():
+    if "LLK_HOME" not in os.environ:
```
fvranicTT, line 21: you could favor early return to avoid deep nesting:
```python
if "LLK_HOME" in os.environ:
    return
```

tests/python_tests/conftest.py
```diff
@@ -17,6 +17,20 @@
...
+        helpers = os.path.abspath(os.path.dirname(__file__))
```
fvranicTT, line 22: you should be able to use __file__ like this:
```python
base_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
os.environ["LLK_HOME"] = base_dir
```

fvranicTT (approved):

tests/python_tests/conftest.py
```diff
@@ -17,6 +17,20 @@
...
+
+        os.environ["LLK_HOME"] = os.path.abspath(
+            os.path.join(
+                helpers,
+                "..",
+                "..",
+                "..",
+            )
+        )
```
skotaracTT, line 31:
```suggestion
def init_llk_home():
   if "LLK_HOME" not in os.environ:
    try:
        helpers = Path(__file__).resolve().parent
        llk_home = helpers.parents[2]  # Up 3 levels
        os.environ["LLK_HOME"] = str(llk_home)
    except Exception as e:
        raise RuntimeError("Failed to set LLK_HOME from script location.") from e
```
Could use pathlib.Path for clearer + reliable path handling. Don't need to hardcode "..",  explicitly navigates directory levels.

PR #393, opened 2025-06-20
[Test Infra] Enable Reading from Dest Register, Supporting Integer formats

skotaracTT:
### Ticket

https://github.com/orgs/tenstorrent/projects/123/views/1?pane=issue&itemId=116276317&issue=tenstorrent%7Ctt-llk%7C392
### Problem description

It’s currently difficult to debug and understand why certain test cases are failing because exalens tool doesn't support reading from src and dest registers when the data format is any type of integer. As a result, running tests with integer format combinations that fail becomes a black box — we get a failure, but not much insight into why.

### What's changed

Generated a wrapper function that will use existing exalens functions to read from the device using TRISC at a given address. Enables printing the dest register contents and is supported for all Tensix integer formats, which should provide some much-needed visibility until native support for integer format reads is implemented.
### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/helpers/device.py
+++ b/tests/python_tests/helpers/device.py
@@ -215,7 +215,7 @@ def read_dest_register(dest_acc: DestAccumulation, num_tiles: int = 1):
     location = RiscLoc(loc=coordinate, noc_id=noc_id, risc_id=risc_id)
     debug_risc = RiscDebug(location=location, context=context, verbose=False)
 
-    assert num_tiles <= 8 if dest_acc == DestAccumulation.Yes else num_tiles <= 16
+    assert num_tiles <= (8 if dest_acc == DestAccumulation.Yes else 16)
 
     word_size = 4  # bytes per 32-bit integer
     num_words = num_tiles * 1024


tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
 
 
+def read_dest_register(num_tiles: int = None, dest_acc: DestAccumulation = None):
+    """
+    Reads values in the destination register from the device.
+    Args:
+        num_tiles: Number of tiles to read from the destination register.
+        dest_acc: Whether destination accumulation is enabled or not.
+
+    Prerquisite: Disable flag that clear dest register after packing (in llk_pack_common.h) otherwise you will read garbage values.
```
nvelickovicTT, line 181: nit: typos; `Prerequisite`, `clears`

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
+        - comment out this line : TT_ZEROACC(p_zeroacc::CLR_HALF, is_fp32_dest_acc_en, 0, ADDR_MOD_1, (dest_offset_id) % 2);
+
+    Note:
+        - The destination register is read from the address 0xFFBD8000.
```
nvelickovicTT, line 185: Is this applicable to both WH and BH?

nvelickovicTT (approved):

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
+        - The number of tiles must be specified if destination accumulation is enabled.
```
fvranicTT, line 186:
```suggestion
        - If destination accumulation is enabled, num_tiles must be ≤ 8. Otherwise, ≤ 16.
```

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
 
+def read_dest_register(num_tiles: int = None, dest_acc: DestAccumulation = None):
+    """
+    Reads values in the destination register from the device.
+    Args:
+        num_tiles: Number of tiles to read from the destination register.
+        dest_acc: Whether destination accumulation is enabled or not.
+
+    Prerquisite: Disable flag that clear dest register after packing (in llk_pack_common.h) otherwise you will read garbage values.
+        - comment out this line : TT_ZEROACC(p_zeroacc::CLR_HALF, is_fp32_dest_acc_en, 0, ADDR_MOD_1, (dest_offset_id) % 2);
```
fvranicTT, line 182:
```suggestion
        - Comment out this line: TT_ZEROACC(p_zeroacc::CLR_HALF, is_fp32_dest_acc_en, 0, ADDR_MOD_1, (dest_offset_id) % 2);
```

tests/python_tests/helpers/device.py
skotaracTT: This is a good point, it differs a bit for wormhole. I'm working on updating this and will add to docstring 👍

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
+        check_context,
+        convert_coordinate,
+        validate_device_id,
+    )
+
+    risc_id = 1  # we want to use TRISC 0 for reading the destination register
+    noc_id = 0  # NOC ID for the device
+    device_id = 0  # Device ID for the device
+    core_loc = "0,0"  # Core location in the format "tile_id,risc_id"
+    base_address = 4290609152
```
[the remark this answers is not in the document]
skotaracTT: https://github.com/tenstorrent/tt-exalens/blob/main/ttexalens/hardware/baby_risc_debug.py#L677

tests/python_tests/helpers/device.py
skotaracTT: That's a good point, this differs for wormhole but it is also not supported on wormhole. Since TRISCs need to be halted in order to read from memory... there is a bug in hardware which causes TRISCs to exit halted state after reading from memory once (note we read from memory in a for loop 1024 times)... This is fine if it happens on blackhole because `debug_risc.read_memory()` will catch this and halt the core again, but on wormhole it is unable to halt any risc core again (ever). Since we have a for loop that reads 32bits from consecutive addresses, after it does the first read and iterates the loop to read again the test will fail because the TRISC is not halted and cannot be halted again. 

Debug team has not provided a workaround for this.
So for now we read from dest on blackhole, and I am unblocked when conducting research on blackhole until debug team provides support for integer reading for both architectures.
I will add this in the docstring.

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
+    validate_device_id(device_id, context)
+    coordinate = convert_coordinate(core_loc, device_id, context)
+
+    if noc_id not in (0, 1):
+        raise ValueError("Invalid value for noc_id. Expected 0 or 1.")
+
+    if risc_id < 0 or risc_id > 3:
+        raise ValueError("Invalid value for risc_id. Expected value between 0 and 3.")
+
+    location = RiscLoc(loc=coordinate, noc_id=noc_id, risc_id=risc_id)
```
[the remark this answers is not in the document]
skotaracTT: Yes, I saw. It's using risc_name now. This function will need to be refactored once we stop docking from the past 😁

tests/python_tests/helpers/device.py
```diff
@@ -171,6 +171,56 @@ def get_result_from_device(
...
+    noc_id = 0  # NOC ID for the device
+    device_id = 0  # Device ID for the device
+    core_loc = "0,0"  # Core location in the format "tile_id,risc_id"
+    base_address = 4290609152
+
+    context = check_context()
+    validate_device_id(device_id, context)
+    coordinate = convert_coordinate(core_loc, device_id, context)
+
+    if noc_id not in (0, 1):
```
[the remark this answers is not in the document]
skotaracTT: Yes, I agree. I had this in mind for the future... We might have DM integration which will look at different NOCs, even though it is hardcoded to one now. 
I will need to keep the assertion for `risc_id == 1` since only TRISC0 can halt successfully during consecutive reads to memory.

PR #404, opened 2025-06-25
profiler ⌛: refactor how metadata is passed

sstanisicTT:
### Problem description
The metadata that is needed for the profiler to properly function is currently passed compile time from `gcc` to the python infra using `#pragma message`. The `stderr` stream is then parsed on the python side for this metadata. This has already caused issues when files containing profiler primitives are not recompiled every run, because the message wouldn't get written to `stderr` 

### What's changed
This refactor changes how profiler metadata is passed to python infra. The profiler metadata will get pushed into the `.profiler_meta` section, which will get parsed after linking is done. Because meta is kept in the intermediate `.o` files, we have a guarantee that all metadata will end up in the final `.elf` of the corresponding thread.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -13,6 +13,7 @@ SHARED_DIR       := $(BUILD_DIR)/shared
 TEST_DIR        := $(BUILD_DIR)/tests/$(testname)
 OBJ_DIR         := $(TEST_DIR)/obj
 DIS_DIR         := $(TEST_DIR)/dis
+PROFILER_DIR    := $(TEST_DIR)/profiler
 ELF_DIR         := $(TEST_DIR)/elf
 
 HELPERS         := helpers
@@ -23,6 +24,7 @@ RMDIR           := rm -rf
 
 GXX             := $(TOOL_PATH)/riscv32-tt-elf-g++
 OBJDUMP         := $(TOOL_PATH)/riscv32-tt-elf-objdump
+OBJCOPY         := $(TOOL_PATH)/riscv32-tt-elf-objcopy
 
 # =========================
 # Architecture Selection
@@ -61,7 +63,7 @@ TO_UPPER = $(shell echo $(1) | tr '[:lower:]' '[:upper:]')
 # =========================
 # Targets
 # =========================
-.PHONY: all dis clean
+.PHONY: all dis profiler clean
 
 all: $(ELF_DIR)/unpack.elf \
 	 $(ELF_DIR)/math.elf \
@@ -73,6 +75,9 @@ dis: $(DIS_DIR)/unpack.S \
 	 $(DIS_DIR)/pack.S \
 	 $(SHARED_DIR)/brisc.S
 
+profiler: $(PROFILER_DIR)/unpack.meta.bin \
+	      $(PROFILER_DIR)/math.meta.bin \
+	      $(PROFILER_DIR)/pack.meta.bin
 # =========================
 # Build Rules
 # =========================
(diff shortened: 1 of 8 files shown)


tests/Makefile
```diff
@@ -73,6 +75,9 @@ dis: $(DIS_DIR)/unpack.S \
 	 $(DIS_DIR)/pack.S \
 	 $(SHARED_DIR)/brisc.S
 
+profiler: $(PROFILER_DIR)/unpack.meta.bin \
+	      $(PROFILER_DIR)/math.meta.bin \
+	      $(PROFILER_DIR)/pack.meta.bin
```
fvranicTT, line 80:
```suggestion
	  $(PROFILER_DIR)/math.meta.bin \
	  $(PROFILER_DIR)/pack.meta.bin
```

tests/Makefile
```diff
@@ -23,6 +24,7 @@ RMDIR           := rm -rf
 
 GXX             := $(TOOL_PATH)/riscv32-tt-elf-g++
 OBJDUMP         := $(TOOL_PATH)/riscv32-tt-elf-objdump
+OBJCOPY         := $(TOOL_PATH)/riscv32-tt-elf-objcopy
```
fvranicTT, line 27: Since you're already touching Makefile, could you add `-g` to line 47? :)

tests/helpers/include/profiler.h
```diff
@@ -36,6 +34,14 @@ constexpr std::uint16_t hashString16(const char (&s)[N])
 
 #define MARKER_ID(marker) hashString16(MARKER_FULL(marker))
 
+#define PROFILER_META(full_marker)                          \
+    {                                                       \
```
fvranicTT, line 38: Can you add a comment here, explaining what this code does in plane English? 😁

tests/helpers/include/profiler.h
```diff
@@ -36,6 +34,14 @@ constexpr std::uint16_t hashString16(const char (&s)[N])
...
+        asm volatile(                                     \
```
fvranicTT, line 39: After another thought, wouldn't this achieve the same thing?
```cpp
#define PROFILER_META_SIMPLE(marker) \
    __attribute__((used, section(".profiler_meta"))) \
    static const char profiler_meta_##__COUNTER__[] = marker;
```
Only it would do it at compile time. And it looks a bit easier to understand IMO.

sstanisicTT: both of these are done at compile time because the directives are evaluated by the assembler. I went with the inline assembly method because it felt more explicit about how it works, but I'll think about it.

tests/Makefile
sstanisicTT: Done ;)

tests/helpers/include/profiler.h
sstanisicTT: Sure

tests/helpers/include/profiler.h
```diff
@@ -36,6 +34,12 @@ constexpr std::uint16_t hashString16(const char (&s)[N])
 
 #define MARKER_ID(marker) hashString16(MARKER_FULL(marker))
 
+/* Push a string containing the full marker into the .profiler_meta section.
+ * This section will be processed by the host code to construct a mapping
+ * MARKER_ID -> { filename, line, marker } for parsing the profiler buffer.
+ */
+#define PROFILER_META(full_marker) __attribute__((section(".profiler_meta"), used)) static const char _profiler_meta_##__COUNTER__[] = full_marker;
```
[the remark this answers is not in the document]
sstanisicTT: clang-format wanted it in a single line, but I can probably turn it off in this case

PR #411, opened 2025-06-26
Enable BRISC TRISC synchronisation

amokanTT:
BRISC handles the soft reset of the TRISCs, which ensures that dst is zeroed properly via TTI_ZEROACC.

### Problem description
When tests are run in simulation, the BRISC does not start executing on time, and therefore the dst register is not zeroed. Do to the lack of synchronisation, TRISC executes and takes garbage values from dst and produces incorrect results.

### What's changed
The problem is solved by introducing BRISC <-> TRISC synchronisation. The BRISC is taken out of reset first, and then the BRISC is responsible for taking the TRISCs out of reset. This ensures that the dst register is zeroed before TRISCs start running.

### Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/helpers/device.py
+++ b/tests/python_tests/helpers/device.py
@@ -60,9 +60,6 @@ def collect_results(
 def run_elf_files(testname, core_loc="0,0", run_brisc=True):
     BUILD = "../build"
 
-    if run_brisc:
-        run_elf(f"{BUILD}/shared/brisc.elf", core_loc, risc_id=0)
-
     context = check_context()
     device = context.devices[0]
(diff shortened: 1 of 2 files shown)


tests/python_tests/helpers/device.py
```diff
@@ -60,9 +60,6 @@ def collect_results(
 def run_elf_files(testname, core_loc="0,0", run_brisc=True):
```
sstanisicTT, line 60: Remove this, because from now on the TRISC cores won't go out of reset if BRISC is not running

PR #422, opened 2025-06-27
Relocate the TRISC kernel running write to mailbox

amokanTT:
Write to the mailbox that the TRISC kernel is running immediately after detecting that tensix is complete, instead of doing it in trisc.cpp

### Problem description
When running multiple tests one after the other on simulator, only the first one passes, and the following ones fail. This is because at the end of the first test there is a 1 written to the mailbox. This value gets read, instead of the 2 which is supposed to be written into the mailbox by trisc.cpp. This results in the second test getting interpreted as already finished, when it has not even started yet. 

### What's changed
Instead of trisc.cpp writing the 2 into the mailbox, it is now done inside of device.py, right after detecting that Tensix operations are complete. 

### Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/conftest.py
+++ b/tests/python_tests/conftest.py
@@ -21,9 +21,6 @@ from helpers.format_arg_mapping import Mailbox
 from helpers.log_utils import _format_log
 from helpers.perf import delete_reports
 
-# Constant - indicates the TRISC kernel run status
-RESET_VAL = 0  # Kernel not running and not complete
-
 
 def init_llk_home():
     if "LLK_HOME" in os.environ:
@@ -69,10 +66,13 @@ def set_chip_architecture():
 
 
 @pytest.fixture(autouse=True)
-def setup_all_mailboxes_before_test():
-    write_words_to_device(core_loc="0, 0", addr=Mailbox.Packer.value, data=RESET_VAL)
-    write_words_to_device(core_loc="0, 0", addr=Mailbox.Math.value, data=RESET_VAL)
-    write_words_to_device(core_loc="0, 0", addr=Mailbox.Unpacker.value, data=RESET_VAL)
+def reset_mailboxes():
+    """Reset all core mailboxes before each test."""
+    core_loc = "0, 0"
+    reset_value = 0  # Constant - indicates the TRISC kernel run status
+    mailboxes = [Mailbox.Packer, Mailbox.Math, Mailbox.Unpacker]
+    for mailbox in mailboxes:
+        write_words_to_device(core_loc=core_loc, addr=mailbox.value, data=reset_value)
     yield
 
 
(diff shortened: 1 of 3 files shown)


ldjurovicTT (approved):

fvranicTT (approved):

tt-vjovanovic:
I don't think this is the way to go.
Image you have a test that is failing or you have exception in python and you don't execute cleanup.
You need to guarantee that test setup prepares everything accordingly.

tests/python_tests/helpers/device.py
```diff
@@ -89,7 +93,9 @@ def run_elf_files(testname, core_loc="0,0"):
     TRISC = ["unpack", "math", "pack"]
     for i in range(3):
         load_elf(
-            elf_file=f"{BUILD}/tests/{testname}/elf/{TRISC[i]}.elf",
+            elf_file=str(
```
fvranicTT, line 96: Looking at the whole method, I think we could refactor it to resemble something like:

```python
def run_elf_files(testname, core_loc="0,0"):
    build_dir = Path("../build")
    
    # Perform soft reset
    perform_tensix_soft_reset(core_loc)
    
    # Load TRISC ELF files
    trisc_names = ["unpack", "math", "pack"]
    for i, trisc_name in enumerate(trisc_names):
        elf_path = build_dir / "tests" / testname / "elf" / f"{trisc_name}.elf"
        load_elf(
            elf_file=str(elf_path.absolute()),
            core_loc=core_loc,
            risc_name=f"trisc{i}",
        )
    
    # Reset the profiler barrier
    TRISC_PROFILER_BARRIER_ADDRESS = 0x16AFF4
    write_words_to_device(core_loc, TRISC_PROFILER_BARRIER_ADDRESS, [0, 0, 0])

    # Run BRISC
    brisc_elf_path = build_dir / "shared" / "brisc.elf"
    run_elf(
        str(brisc_elf_path.absolute()), 
        core_loc, 
        risc_name="brisc"
    ) 
```

tests/python_tests/conftest.py
```diff
@@ -61,6 +68,14 @@ def _identify_chip_architecture(output):
     return architecture
 
 
+@pytest.fixture(autouse=True)
```
fvranicTT, line 71: Also, this could be written in a way that avoids repetition:
```python
@pytest.fixture(autouse=True)
def reset_mailboxes():
    """Reset all core mailboxes before each test."""
    core_loc = "0, 0"  # Could be parameterized if needed
    # Constant - indicates the TRISC kernel run status
    reset_value = 0
    mailboxes = [Mailbox.Packer, Mailbox.Math, Mailbox.Unpacker]
    for mailbox in mailboxes:
        write_words_to_device(core_loc=core_loc, addr=mailbox.value, data=reset_value)
    
    yield
```
In that case, you don't need `RESET_VAL` from above.

tt-vjovanovic (approved):

PR #444, opened 2025-07-01
perf :zap:: generate baseline scatter graph

sstanisicTT:
### Problem description
In a future PR the code

### What's changed
Generate a Scatter graph for the performance benchmark

image

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ compile_commands.json
 .cache/**
 .envrc
 .vscode/
-perf/
+perf_data/
 *.log
 *.csv
 *.whl
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ compile_commands.json
 .cache/**
 .envrc
 .vscode/
+perf/
 *.log
 *.csv
 *.whl
--- a/tests/python_tests/helpers/perf.py
+++ b/tests/python_tests/helpers/perf.py
@@ -233,7 +233,7 @@ def delete_benchmark_dir(testname: str):
     if not root:
         raise AssertionError("Environment variable LLK_HOME is not set")
 
-    path = Path(root) / "perf" / testname
+    path = Path(root) / "perf_data" / testname
 
     if path.exists() and path.is_dir():
         shutil.rmtree(path)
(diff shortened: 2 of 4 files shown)


fvranicTT:
Just a note: this scatter plot is overwhelming. I literally can't see the tree from the forest 😉

sstanisicTT:
The plot is definitely overwhelming, but if you filter by run type it's much more manageable. It also shows the sweep params on hover to make it easier to interpret the data:

image

The idea is to have this general graph for all benchmarks, and the ability to generate more specific ones on a benchmark by benchmark basis

tests/python_tests/perf_eltwise_binary_fpu.py
```diff
@@ -50,6 +51,7 @@ def report_fixture():
     delete_benchmark_dir(TEST_NAME)
     yield
     dump_report(TEST_NAME, report)
+    dump_scatter(TEST_NAME, report)
```
fvranicTT, line 54: do we maybe want to make this step optional? i.e. to pass an option to pytest when we would like to have a scatter plots, so we could speed up the test execution? Namely, in CI runs, we don't need this step.

.gitignore
```diff
@@ -11,6 +11,7 @@ compile_commands.json
 .cache/**
 .envrc
 .vscode/
+perf/
```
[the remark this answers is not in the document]
sstanisicTT: this folder is supposed to contain all of the benchmarking data, not just plots, but maybe we can rename it to `perf_data`

tests/python_tests/helpers/perf.py
```diff
@@ -270,3 +272,57 @@ def dump_report(testname: str, report: PerfReport):
...
+
+    mean_columns = [
+        (name, i) for i, name in enumerate(report.stat_names) if name.startswith("mean")
+    ]
+
+    # For each stat column (run type), plot all points
+    for stat_name, stat_idx in mean_columns:
+        y_vals = [stat[stat_idx] for stat in report.stat_values]
+
+        hover = [
```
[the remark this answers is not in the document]
sstanisicTT: Nice catch

tests/python_tests/perf_eltwise_binary_fpu.py
sstanisicTT: We should probably create a separate Issue/PR for CI support and related changes. I think the issue is much deeper because we are using multiple runners for the tests.

fvranicTT (approved):

tests/python_tests/perf_eltwise_binary_fpu.py
sstanisicTT: Created #446

PR #508, opened 2025-07-16
ci: fix permissions of check-stale workflow

fvranicTT:
### Ticket
None

### Problem description
We have some stale PRs that we'd like to automatically label as stale. This workflow has been added some time ago, but after moving to GH enterprise, it stopped working as expected since it was missing permissions.

### What's changed
- Added missing permissions to the check-stale workflow.
- Increased number of days before issue becomes stale to 700, since I wasn't sure about the number already in place (100 days)

### Type of change

- [ ] Documentation update

### DIFF
--- a/.github/workflows/check-stale.yml
+++ b/.github/workflows/check-stale.yml
@@ -3,7 +3,6 @@ name: Close stale issues and PRs
 on:
   schedule:
     - cron: '0 3 * * *'  # Runs daily at 3:00 UTC
-  workflow_dispatch:
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-stale.yml
```diff
@@ -3,16 +3,21 @@ name: Close stale issues and PRs
...
+      contents: read
+      actions: write
+      issues: write
+      pull-requests: write
     steps:
       - uses: actions/stale@v9
         with:
           # Issue settings
-          days-before-stale: 100
+          days-before-stale: 700
```
nvelickovicTT, line 20: not sure how useful this limit will be. 2+ years is really a long time to wait.

nvelickovicTT (approved):

.github/workflows/check-stale.yml
Aswinmcw: 6 months? metal repo is using 6 i think

fvranicTT: I set it to 90 days for PRs and 180 days for issues. We can adjust that if it is not enough.

PR #537, opened 2025-07-25
Inital AI generation of test cases

ldjurovicTT:
### Ticket
None

### Problem description
For the most part LLK tests are handwritten and then ran and debugged. The idea is to utilize existing LLK tests and feed them to AI that will try to generate a lot more tests in much faster time than engineer.

### What's changed
Added ai_gen/ folder to both ptyhon_tests/ and sources/ which holds python and cpp tests respectively. Also Makefile had to be slightly modified in order to support that new path that emerged.

#### Prompting strategy:

The AI was given prompts like: `Analyze existing tests in test/ folder and then having in mind how they are written write a test that will have unpack_A, eltwise, sfpu and regular pack. Also use only Float16_b format. Sweep 32 bit dest acc, fidelity, dest sync.`
If it generated code that cannot be compiled tool was just given error output from terminal. In every case it managed to solve build errors.
When test are failing model was given test outputs( golden and result tensors ) and given task to analyze them and correct test. It is much harder then build errors and not many times did it succeed without human change in tests.

#### Observation
This AI generation method can be very useful when adding new parameters for sweeps in full API sweep tasks. It can easily detect where to add parameters and how to use them. 
Further usage could be to feed the tool with these AI generated fused tests and ask it to make fuser which would allow user to specify which APIs to fuse and autogenerates everything else.

#### Current state:
Tool generated 3 python and 3 cpp tests which are:
- unpack_AB, eltwise binary, sfpu, pack untilize
- unpack_AB, eltwise binary, sfpu, pack
- unpack_AB, reduce, sfpu, pack

image

### Type of change

- [ ] Documentation update


tests/helpers/include/ckernel_helper.h
```diff
@@ -6,7 +6,11 @@
 
 #include <cstdint>
 
-constexpr std::uint32_t INSTRN_BUF_BASE = 0xFFE40000;
+#ifdef INSTRN_BUF_BASE
+#undef INSTRN_BUF_BASE
+#endif
+
+#define INSTRN_BUF_BASE 0xFFE40000
```
fvranicTT, line 13: We shouldn't be needing this anymore, right? After Nathan made a fix a few days ago, this should be infered automatically.

tests/python_tests/helpers/format_arg_mapping.py
```diff
@@ -180,6 +180,13 @@ class Mailbox(Enum):
 }
 
 
+class DstSync(Enum):
```
fvranicTT, line 183:
```suggestion
class DestSynchronization(Enum):
```
No need to be shy with names. It is easier to read when you're new and looking at the code for the first time.

tests/python_tests/helpers/golden_generators.py
```diff
@@ -364,8 +364,11 @@ def __init__(self):
         }
         self.data_format = None
 
-    def __call__(self, operation, operand1, data_format):
+    def __call__(
+        self, operation, operand1, data_format, math_fidelity=MathFidelity.LoFi
+    ):
         self.data_format = data_format
+        self._math_fidelity = math_fidelity
```
fvranicTT, line 371: Should we use _ for class internal members? In that case, data format should follow suite.

tests/helpers/include/sfpu_operations.h
```diff
@@ -0,0 +1,129 @@
...
+            break;
+        default:
+            return; // Unsupported op – should never happen
+    }
+}
+
+/**
+ * Convenience function for the common case of 32 iterations (full tile)
+ */
+inline void call_sfpu_operation_32(SfpuType operation, uint32_t math_format = 0)
```
fvranicTT, line 124: not easy to understand when read in a different file what 32 is.

tests/sources/ai_gen/eltwise_binary_sfpu_pack_untilize.cpp
```diff
@@ -0,0 +1,161 @@
...
+    _llk_math_eltwise_binary_<ELTWISE_BINARY_OP, BroadcastType::NONE, DST_SYNC, is_fp32_dest_acc_en, MATH_FIDELITY, EltwiseBinaryReuseDestType::NONE>(
+        4, 0, false);
+
+    // SFPU unary on result in dest
+    _llk_math_eltwise_unary_sfpu_init_<SFPU_UNARY_OPERATION>();
+    _llk_math_eltwise_unary_sfpu_start_<DST_SYNC>(0);
+
+    constexpr int iterations = 32;
+
+    auto call_sfpu_operation = [](SfpuType op)
```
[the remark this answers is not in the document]
ldjurovicTT: done

PR #558, opened 2025-08-01
Change ALU_FORMAT of SrcA in reduce init to fix MOVB2D bug with Tf32

rdjogoTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/25266

### Problem description
Setting Tf32 as ALU FORMAT for SrcA changes behaviour of MOVB2D/D2B. This arises in situations where the values in SrcA (being reduced) are Tf32.

This only affects WH, BH works correctly for Tf32 as well.

### What's changed
Set to valid format in the reduce init, since expecting OP writers to know this HW quirk seems unreasonable.
This must be accompanied by tt-metal change as well. (https://github.com/tenstorrent/tt-metal/pull/26528)

### Type of change

- [ ] Documentation update

### Checklist
- [x] [All post commit](https://github.com/tenstorrent/tt-metal/actions/runs/16833535686) CI passes
- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/runs/16833537470) CI with demo tests passes (if applicable)

### DIFF
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
@@ -436,7 +436,7 @@ inline void reduce_configure_mop()
     }
 }
 
-template <PoolType type, ReduceDim dim, int MATH_FIDELITY_DESC = 0>
+template <PoolType type, ReduceDim dim, int MATH_FIDELITY_DESC = 0, bool fp32_transpose = false>
 inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpose = 0)
 { // within_face_16x16_transpose used for unpack, ignored by math
 
@@ -449,6 +449,11 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
         reduce_configure_mop<dim, MATH_FIDELITY_PHASES>();
     }
 
+    if constexpr (fp32_transpose)
+    {
+        // MOVB2D depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32
+        cfg_reg_rmw_tensix<ALU_FORMAT_SPEC_REG0_SrcA_RMW>((uint)DataFormat::Float16_b);
+    }
     TTI_SETC16(CLR_DVALID_SrcA_Disable_ADDR32, 0);
 
     math::reset_counters(p_setrwc::SET_ABD_F);
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
@@ -460,7 +460,7 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
         static_assert(is_fp32_dest_acc_en, "FP32 Dest must be enabled for FP32 accumulation");
         // MOVB2D/D2B depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32 (only on WH)
         // This is needed because FP32 data from L1 that is unpacked to Src registers is reduced to Tf32
-        cfg_reg_rmw_tensix<ALU_FORMAT_SPEC_REG0_SrcA_RMW>((uint)DataFormat::Float16_b);
+        cfg_reg_rmw_tensix<ALU_FORMAT_SPEC_REG0_SrcA_RMW>((uint)DataFormat::Float32);
     }
     TTI_SETC16(CLR_DVALID_SrcA_Disable_ADDR32, 0);
 
--- a/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
@@ -451,7 +451,7 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
 
     if constexpr (fp32_transpose)
     {
-        // MOVB2D depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32
+        // MOVB2D depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32 (only on WH)
         cfg_reg_rmw_tensix<ALU_FORMAT_SPEC_REG0_SrcA_RMW>((uint)DataFormat::Float16_b);
     }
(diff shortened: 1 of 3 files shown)


tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
```diff
@@ -449,6 +449,11 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
         reduce_configure_mop<dim, MATH_FIDELITY_PHASES>();
     }
 
+    if constexpr (fp32_transpose)
+    {
+        // MOVB2D depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32
```
amahmudTT, line 454: 1. You can specify in the comment that it is a wormhole specific thing. 

2. So your branch depends on the added template variable. I am getting a bit confused here, how would the solution look like if the template variable was not added, and is everybody going to use your template variable ?

3. (commenting here although not exactly relevant) has the post commit tests passed.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
```diff
@@ -436,7 +436,7 @@ inline void reduce_configure_mop()
     }
 }
 
-template <PoolType type, ReduceDim dim, int MATH_FIDELITY_DESC = 0>
+template <PoolType type, ReduceDim dim, int MATH_FIDELITY_DESC = 0, bool fp32_transpose = false>
```
nvelickovicTT, line 439: Fp32_transpose is a bit confusing for me.
Is this true: fp32_transpose = (unpack_src_format == DataFormat::FP32).
If so, than I would propose we pass the format here and do the check internally, instead of requiring users to ensure consistency between this flag and unpack_src_format.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
```diff
@@ -449,6 +449,11 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
         reduce_configure_mop<dim, MATH_FIDELITY_PHASES>();
     }
 
+    if constexpr (fp32_transpose)
+    {
+        // MOVB2D depends on SrcA ALU Format - Hi/Lo16 does not work with Tf32 (only on WH)
```
nvelickovicTT, line 454: Maybe just add few words to connect fp32 and tf32.
It may not be immediately obvious to readers that  fp32 input will result in tf32 format on Src register.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
```diff
@@ -449,6 +449,11 @@ inline void _llk_math_reduce_init_(const std::uint32_t within_face_16x16_transpo
...
+        cfg_reg_rmw_tensix<ALU_FORMAT_SPEC_REG0_SrcA_RMW>((uint)DataFormat::Float16_b);
```
jasondavies, line 455: From my reading of the [WH ISA docs](https://github.com/tenstorrent/tt-isa-documentation/blob/60f4906ae5964eae4974c3a5b70667185f62ea24/WormholeB0/TensixTile/TensixCoprocessor/MOVD2B.md#L34-L43), `DataFormat::Float32` should have the same effect.  I don't know if that might be preferable as it aligns more with the actual data format.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
rdjogoTT: Sorry for the delay @nvelickovicTT 

`fp32_transpose` is an alternate mode of operation for the reduce kernel. Either it accumulated in tf32 (if it's set to false) or full fp32 (if true): https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/include/compute_kernel_api/reduce.h

`fp32_transpose = (unpack_src_format == DataFormat::FP32)` is not true, because it depends on whether the OP writer want to do reduction accumulation in fp32 or they want the old approach.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
rdjogoTT: I also haven't been able to find a difference between the two formats, and I have confirmed they both solve the issue (only Tf32 doesn't work).
This makes sense to me, so I'll make the change unless any of the reviewers object @nvelickovicTT @amahmudTT ?

nvelickovicTT: Makes sense to me.

tt_llk_wormhole_b0/llk_lib/llk_math_reduce.h
nvelickovicTT: OK, I see. Sorry, I haven't looked at the implementation deep enough.
So, I understand we need it as a separate flag, however, I still find the name confusing because to understand it requires to know details of the implementation.
Param description for it says: "Enable accumulation of reduction in full FP32 precision". No transpose is mentioned.

What do you think about `enforce_fp32_accumulation`?
And also add to the description that it is only relevant when DataFormat::FP32 is requested.

rdjogoTT: I think that naming makes sense.
But I'm not sure what you mean by the last sentence, "only relevant when DataFormat::FP32 is requested". We can still accumulate in fp32 if the input is in bfp8_b or fp16_b.

nvelickovicTT: I think I still have some gaps in understanding the feature then.
I am under the impression that this flag is there to ensure we do not loose precision when the requested input format is 32bit while doing transpose using SrcB.
If the input format is 16bit, what would be the benefit of using this flag?

rdjogoTT: Yes that is a good point, but I can add a static assert for that and add that info to the comment.

nvelickovicTT: Sounds good. Thanks.

amahmudTT (approved):

fvranicTT (approved):

PR #614, opened 2025-08-25
Use .ttinsn for quasar TTI insns

nathan-TT:
### Ticket
NA

### Problem description
Quasar's `ckernel_ops.h` hasn't been updated to use `.ttinsn`

### What's changed
Update it to match wormhole & blackhole.  Also noticed we should be using `"n"` constraint, not `"I"`.

### Type of change

- [YES] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_quasar/common/inc/ckernel_ops.h
+++ b/tt_llk_quasar/common/inc/ckernel_ops.h
@@ -3,38 +3,35 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //
-// Auto-generated file, do not modify!
+// Modified auto-generated file
 //
 
 #pragma once
 
 #define TT_OP(opcode, params) ((opcode << 24) + params)
-#define INSTRUCTION_WORD(x)   __asm__ __volatile__(".word (%0)" : : "i"((x))) // Drop 32 bits into the instruction stream.
-#define TRISC_OP_SWIZZLE(x) \
-    ((((x) >> 30) & 0x3) |  \
-     (((x) & 0x3FFFFFFF) << 2)) // Put top 2 bits, which are currently never 'b11 to bottom, indicating to Risc that they are not risc instructions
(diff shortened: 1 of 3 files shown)


tt_llk_quasar/common/inc/ckernel_ops.h
```diff
@@ -3,60 +3,55 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //
-// Auto-generated file, do not modify!
+// Modified auto-generated file
```
amahmudTT, line 6: Do we want this comment ?

nathan-TT: I don't know, but the current comment is wrong -- would you prefer no comment? BTW when I modified wormhole & blackhole variants, I tried, but failed, to track down whatever script created these.

nvelickovicTT: @rtawfik01, @rdjogoTT historically, do you have an idea where was this file autogenerated from? @rtawfik01 for the quasar version, it is still fresh, so you might still remember.

nvelickovicTT (approved):
@nathan-TT doing an overkill here and added a metal post commit tests label, that triggers post commit in metal with this change.

tt_llk_quasar/common/inc/ckernel_ops.h
rtawfik01: Yes this gets generated by an rdl file in HW repo, it gets generated after RTL is frozen and then it remains consistent for the remainder of the arch

rtawfik01: Also @nathan-TT please do not modify quasar files while we still do not have regressions for it

nathan-TT:
Also @nathan-TT please do not modify quasar files while we still do not have regressions for it

PR #622, opened 2025-08-27
qsr: cleanup ckernel_risc_atomics.h

sstanisicTT:
### What's changed

Made a few improvements to `load_acquire` and `store_release`:
- They no longer take a pointer to an atomic, but a raw pointer (if it's atomic you can already use std::atomic<T>::load())
- From what I know the volatile cast for `addr` is redundant here
- Switched  `T ret -> uint32_t ret` so that structs can be used as `T`
- Added static assert to check if the underlying type can be atomically loaded

Example:
```cpp
typedef struct fifo {
    uint16_t rd_ptr;
    uint16_t wr_ptr;
} fifo_t;


int main () {
    fifo_t fifo = load_acquire((fifo_t*)0x1000);        // -> works fine
    uint8_t val = load_acquire((uint8_t*)0x2000);  // -> errors on the static assert
    return 0;
}
```

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_quasar/common/inc/ckernel_risc_atomics.h
+++ b/tt_llk_quasar/common/inc/ckernel_risc_atomics.h
@@ -4,107 +4,62 @@
 
 #pragma once
 
-#include <atomic>
 #include <cstdint>
+#include <cstring>
+#include <type_traits>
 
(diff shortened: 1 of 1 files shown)


skotaracTT (approved):
Looks good

Issue #580, opened 2025-08-08
SFPU binary/ternary operations mix the order of operands


rdjogoTT:
In `llk_math_eltwise_binary_sfpu_params.h` (same with ternary), we take the operand Dst indices and use the `min` of them to program the Dst offset for the LLK:

https://github.com/tenstorrent/tt-llk/blob/03872ac9c125f8df4fce44f3fa2e2cb55e220fdf/tt_llk_blackhole/llk_lib/llk_math_eltwise_binary_sfpu_params.h#L14-L16

The offset between the operands is also calculated and passed to the llk as the location of the 2nd operand.

In cases where the index of operand0 <= index of operand1, this is fine, but if that is not true then the operand order will get flipped. This results in incorrect behaviour for non-commutative operations, like `div` or `sub` (ie. doesn't affect `add` or `mul`).

This needs to be addressed by preserving the order of operands when they get passed down to the LLK.

Also, it is worth considering to add the ability to overwrite any selected tile in Dest, rather than just the first operand. This would give more flexibility to OP writers in tt-metal.

fvranicTT:
Can we identify all the affected functions? In that case, this could become a bounty, correct? Provided we know the breadth and depth of the issue.

cmaryanTT:
This is now affecting OPs deliverables for @bbradelTT . Raising this in the priority list.

PR #623, opened 2025-08-27
Change binary/ternary SFPU API and kernels to use tile idx directly

rdjogoTT:
### Ticket
https://github.com/tenstorrent/tt-llk/issues/580
https://github.com/tenstorrent/tt-metal/issues/19442

### Problem description
We used to take the min(tile_idx0, tile_idx1, ...) and program the Dest math offset to that in order to perform SFPU kernels by passing down just the "offset" between operands. This broke when the second operand was located in an earlier tile_idx than the first for non-commutative OPs. Also, the tile at offset 0 was always used as output with no flexibility.

### What's changed
For both BH and WH:
- change the API and kernels to accept an idx for each tile used (input tiles and output tile) and update the kernels as needed.
- where OP (only ternary op) was updated for WH to use TT_ instructions 100% (rather than mixing with sfpi) because when switching to a TT_STORE call it was causing incorrect outputs. Simplest way to unblock was to implement the v_ifs as TT instructions manually.
- I've tested a few OPs (one int32 and one fp32) to check there is not perf loss for switching to TT instructions (from TTI) and results show no change in kernel duration. This makes sense it the compiler does indeed detect that the operands are constexprs at compile time and optimize out the difference

### Type of change

- [x] Documentation update (please point out if I missed any documentation changes)

### Checklist
- [x] [All post commit](https://github.com/tenstorrent/tt-metal/actions/runs/17275851094) CI passes
- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/runs/17275853472) CI with demo tests passes (if applicable)

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_add_int.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_add_int.h
@@ -30,6 +30,7 @@ inline void _add_int_(const uint dst_index_in0, const uint dst_index_in1, const 
     // If LOAD/STORE have the value in INT sign-magnitude format and SFPU needs it as 2's complement.
     constexpr auto INSTR_MOD_CAST = InstrModCast::INT_SIGN_MAGN_TO_INT32_2S_COMP;
 
+    // size of each tile in Dest is 64 rows
     constexpr uint dst_tile_size = 64;
 
 #pragma GCC unroll 8
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_log.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_log.h
@@ -15,6 +15,7 @@ namespace sfpu
 template <bool HAS_BASE_SCALING>
 sfpi_inline void _calculate_log_body_(const uint log_base_scale_factor, const uint dst_idx = 0)
 {
+    // size of each tile in Dest is 64/SFP_DESTREG_STRIDE = 32 rows when using sfpi to load/store
     constexpr uint dst_tile_size_sfpi = 32;
 
     ////////////////////////////
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_where.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_where.h
@@ -13,6 +13,7 @@ namespace ckernel::sfpu
 template <bool APPROXIMATION_MODE, int ITERATIONS>
 inline void _calculate_where_fp16_b_(const uint dst_index_in0, const uint dst_index_in1, const uint dst_index_in2, const uint dst_index_out)
 {
+    // size of each tile in Dest is 64 rows
     constexpr uint dst_tile_size = 64;
 
     for (int i = 0; i < ITERATIONS; i++)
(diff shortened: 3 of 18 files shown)


tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_where.h
```diff
@@ -10,59 +10,47 @@
...
+inline void _calculate_where_fp16_b_(const uint dst_index_in0, const uint dst_index_in1, const uint dst_index_in2, const uint dst_index_out)
 {
     constexpr uint dst_tile_size_rows = 64;
 
-    sfpi::vFloat cond = sfpi::dst_reg[0];
-
     for (int i = 0; i < ITERATIONS; i++)
     {
-        cond = sfpi::dst_reg[0];
+        // load conditional value
```
nvelickovicTT, line 20: @rdjogoTT I'm not sure I understood if you tested locally this change, or are you relying on the CI?
In case of the latter, @ldjurovicTT is this covered in CI, or is there some test to run locally?

nsmithtt:
> @nsmithtt FYI. We discussed these dest indexes in the LLK/ d2m sync. This PR doesn't cover unary SFPU API, but we can cover it in a follow up PR.

This is great! Thank you so much

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_add_int.h
```diff
@@ -17,7 +17,7 @@ namespace sfpu
 {
 
 template <bool APPROXIMATION_MODE, int ITERATIONS, InstrModLoadStore INSTRUCTION_MODE, bool SIGN_MAGNITUDE_FORMAT>
-inline void _add_int_(const uint dst_offset)
+inline void _add_int_(const uint dst_index_in0, const uint dst_index_in1, const uint dst_index_out)
```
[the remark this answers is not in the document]
rdjogoTT: I can't think of any combination that wouldn't make sense, so I don't think we need to worry about this.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_log.h
```diff
@@ -13,12 +13,14 @@ namespace sfpu
 {
 
 template <bool HAS_BASE_SCALING>
-sfpi_inline void _calculate_log_body_(const uint log_base_scale_factor)
+sfpi_inline void _calculate_log_body_(const uint log_base_scale_factor, const uint dst_idx = 0)
 {
+    constexpr uint dst_tile_size = 32;
```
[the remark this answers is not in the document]
rdjogoTT: When using sfpi to load/store, it will automatically multiply the addr by 2 (SFP_DESTREG_STRIDE) so a tile ends up being 32 rows rather than 64 with TTI instructions.
I will rename the cases where it needs to be 32 to dst_tile_size_sfpi for consistency and add a comment as you requested in another comment

tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_where.h
rdjogoTT: I tested both BH and WH using both tt-metal tests and the llk infra tests. All the tt-metal tests also run in CI, I assume the same for our own infra

rdjogoTT: @ldjurovicTT could you confirm they run in out CI, also please review

PR #627, opened 2025-08-29
Fix Pack Untilize on Blackhole

wransom-TT:
### Ticket
N/A

### Problem description
- Pack untilize was not functioning properly on Blackhole when called with a non-zero DST index.

### What's changed
- I updated the MOP config and TT_SET calls have been adjusted to account for the DST index.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_pack_untilize.h
+++ b/tt_llk_blackhole/llk_lib/llk_pack_untilize.h
@@ -201,6 +201,7 @@ inline void _llk_pack_untilize_(
     TT_SETADCZW(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 zw counters
     TT_SETADCXY(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 xy counters
 
+    // Needs to be revisited for perf impact with https://github.com/tenstorrent/tt-llk/issues/632
     // If starting_tile_dst_offset is non-zero, reconfigure the template with the correct offset
     if constexpr (tile_dst_ct_offset != 0)
     {


rtawfik01:
@pmilenkovicTT can you please review this change?

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -196,6 +200,12 @@ inline void _llk_pack_untilize_(
     TT_SETADCZW(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 zw counters
     TT_SETADCXY(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 xy counters
 
+    // If tile_dst_offset is non-zero, reconfigure the template with the correct offset
+    if (tile_dst_offset != 0)
```
nvelickovicTT, line 204: This has a significant perf impact, as for non-zero dest offset, MOp will be reconfigured on every call.
Depending on how often you actually expect this value to change, it might make sense to move the parameter to init call, and do the MOP reconfig there.

wransom-TT: I will need to iterate back and forth between DST 0 and DST 2 indexes repeatedly in my compute kernel so I don't think this call can go in the init function, what do you think?

nvelickovicTT: In that case, let’s see how to avoid the MOP reconfig. This idx change should just result in W counter update.
But for that to be possible, MOP needs to change.
@pmilenkovicTT I know you’ve been looking into it. Let us know if you have any ideas.

wransom-TT: Ok if possible would it be okay to do this optimization as a follow on?

I believe with the most recent commits I just pushed the new MOP config update should never run for existing code.  It will only run for the new max pool with indices feature I am implementing here:

https://github.com/tenstorrent/tt-metal/pull/27138

I am currently unable to run CI for this PR until the changes in this PR are merged.  So it would be ideal if we could merge this functionality to enable the necessary feature without any performance impact to code outside the above referenced PR.

We need the features in the above PR functional by Friday but performance improvements can wait until next month so we should have time to come up with a better solution than this MOP config one.

tt_llk_blackhole/llk_lib/llk_pack_untilize.h
```diff
@@ -196,6 +201,12 @@ inline void _llk_pack_untilize_(
     TT_SETADCZW(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 zw counters
     TT_SETADCXY(p_setadc::PAC, 0, 0, 0, 0, 0b0011); // reset ch0 xy counters
 
+    // If starting_tile_dst_offset is non-zero, reconfigure the template with the correct offset
+    if constexpr (tile_dst_ct_offset != 0)
```
nvelickovicTT, line 205: Please just add a comment with a link to an issue to follow up on this.

nvelickovicTT (approved):

PR #633, opened 2025-09-03
fix: use improved linker scripts and crt file

fvranicTT:
### Type of change

- [ ] Documentation update

### Checklist


tests/helpers/src/brisc.cpp
```diff
@@ -21,12 +21,4 @@ int main()
     // Release reset of triscs here in order to achieve brisc <-> trisc synchronization
     clear_trisc_soft_reset();
 #endif
-
-    // Use a volatile variable to prevent the compiler from optimizing away the loop
-    volatile bool run = true;
```
fvranicTT, line 26: Note to reviewers: no need for the loop in brisc/trisc.cpp, as we now have infinite loop in the crt file?

tests/helpers/tmu-crt0.S
```diff
@@ -147,7 +149,15 @@ address_fix:
...
-  tail    exit
+
+  # Automatically dump gcda into L1 when main returns if built with coverage.
+  #ifdef COVERAGE
+  call gcov_dump
+  #endif
+
+  # Loop infinitely.
+  infinite_loop: j infinite_loop
+
```
sstanisicTT, line 160: This should a local symbol `.Linfinite_loop` so that it can't collide with a function named `infinite_loop` in linking. Or just use the number based labels, because they are already local

tests/helpers/ld/memory.wormhole.ld
```diff
@@ -4,32 +4,74 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+L0_BASE = 0xFFB00000;
+BRISC_L0_LENGTH = 4K;
+TRISC_L0_LENGTH = 2K;
+NCRISC_L0_LENGTH = 4K;
```
sstanisicTT, line 10: I think `LOCAL_MEM` is more fitting then L0?

fvranicTT: Done.

nvelickovicTT (approved):

PR #654, opened 2025-09-13
chore: use stricter rules to force compiler to respect C++17 standard version

fvranicTT:
### Ticket
None

### Problem description
This PR removes C++ designated initializers from the codebase and replaces them with C++17-compliant builder pattern or brace initialization to ensure compatibility with stricter C++17 compiler rules. 
This should now make the LLKs compliant with C++17 in full.

### What's changed
- Replacement of designated initializers like {.field = value} with builder pattern usage
- Addition of fluent builder classes for address modifier configuration
- Conversion of unions with designated initializers to brace initialization {0}

### Type of change

- [ ] Documentation update


tests/firmware/riscv/common/risc_attribs.h
```diff
@@ -7,14 +7,17 @@
...
     {
-        uint32_t hi;
-        uint32_t lo;
+        uint64_t v;
+
+        struct
+        {
+            uint32_t hi;
+            uint32_t lo;
+        } parts;
```
sstanisicTT, line 20: I'm fairly confident we don't need this. I don't see a reason why the compiler wouldn't be able to just dereference the `uint64_t*` into 2 `lw` instructions. We might be able to just get away with dereferencing the pointer to L1 without this logic with unions.

tt_llk_blackhole/common/inc/ckernel_addrmod.h
```diff
@@ -143,6 +143,144 @@ struct addr_mod_t
...
+// Fluent builder for addr_mod_t - provides clear, explicit parameter setting
+class addr_mod_builder
+{
+private:
+    addr_mod_t mod {};
+
+public:
+    constexpr addr_mod_builder& srca_incr(uint8_t val)
+    {
+        mod.srca.incr = val;
```
sstanisicTT, line 155: We might want to consider giving these friendlier names, just so that people don't have problems figuring out what the shorthands mean

sstanisicTT (approved):

tests/firmware/riscv/common/risc_attribs.h
fvranicTT: We can maybe remove this abomination all together, since it is unused. We still need to keep definitions of `tt_l1_ptr` and `tt_reg_ptr`, as these are in use. WDYT?

nvelickovicTT (approved):

PR #656, opened 2025-09-15
profiler + perf :zap:: port to pandas

sstanisicTT:
### Problem description

To make analysis of the performance benchmark data easier, I will be porting most of the logic in the suite to Pandas. This will make it easier to create reports, do statistical analysis etc. 

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/helpers/profiler.py
+++ b/tests/python_tests/helpers/profiler.py
@@ -2,19 +2,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import csv
 import os
 import re
-from dataclasses import dataclass, fields
+from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Optional
 
+import pandas as pd
 from ttexalens.tt_exalens_lib import read_words_from_device
 
 from helpers.chip_architecture import get_chip_architecture
-from helpers.test_config import ProfilerBuild, generate_make_command
-from helpers.utils import run_shell_command
 
 
 @dataclass
@@ -23,37 +20,6 @@ class ProfilerFullMarker:
     file: str
     line: int
     id: int
-
-
-def build_with_profiler(test_config):
-    make_cmd = generate_make_command(test_config, ProfilerBuild.Yes)
-    run_shell_command(f"cd .. && {make_cmd}")
-
-
-@dataclass
-class ProfilerTimestamp:
-    full_marker: ProfilerFullMarker
-    timestamp: int
-    data: Optional[int] = None
-
-
-@dataclass
-class ProfilerZoneScoped:
-    full_marker: ProfilerFullMarker
-    start: int
-    end: int
-    duration: int
-
-
-@dataclass
-class ProfilerData:
-    unpack: list[ProfilerTimestamp | ProfilerZoneScoped]
-    math: list[ProfilerTimestamp | ProfilerZoneScoped]
-    pack: list[ProfilerTimestamp | ProfilerZoneScoped]
-
-    def __iter__(self):
-        for field in fields(self):
-            yield field.name, getattr(self, field.name)
 
 
 class Profiler:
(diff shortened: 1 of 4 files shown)


tests/python_tests/test_profiler_primitives.py
```diff
@@ -31,51 +32,75 @@ def test_profiler_primitives():
...
+    ), f"Expected timestamp > 0, got {timestamp['timestamp']}"
+    assert pd.isna(
+        timestamp["duration"]
+    ), f"Expected duration to be None/NaN, got {timestamp['duration']}"
+    assert pd.isna(
+        timestamp["data"]
+    ), f"Expected timestamp data to be None/NaN, got {timestamp['data']}"
+
+    # TIMESTAMP_DATA - Get first TIMESTAMP type entry from PACK thread with data
+    data_timestamps = df[
```
fvranicTT, line 83:
```python
data_timestamps = df[
    (df["thread"] == "PACK") & 
    (df["type"] == "TIMESTAMP") & 
    (df["marker"] == "TEST_TIMESTAMP_DATA")
]
```

Or since this is now repeating a few times, we could make it into a utility function:
```python
def get_filtered_data(df, thread, type, marker):
    return df[
        (df["thread"] == thread) & 
        (df["type"] == type) & 
        (df["marker"] == marker)
    ]

zones = get_filtered_data(df, "UNPACK", "ZONE", "TEST_ZONE")
timestamps = get_filtered_data(df, "MATH", "TIMESTAMP", "TEST_TIMESTAMP")
data_timestamps = get_filtered_data(df, "PACK", "TIMESTAMP", "TEST_TIMESTAMP_DATA")
```

sstanisicTT: I will probably try to make a wrapper class around pandas DataFrame that will allow me to easily do specific filters like this one

fvranicTT (approved):

tests/python_tests/helpers/profiler.py
```diff
@@ -25,35 +22,40 @@ class ProfilerFullMarker:
...
+        """Return the underlying DataFrame"""
+        return self.df
 
-@dataclass
-class ProfilerTimestamp:
-    full_marker: ProfilerFullMarker
-    timestamp: int
-    data: Optional[int] = None
+    # Filter by thread
+    def unpack(self) -> "ProfilerData":
```
fvranicTT, line 34: I like this 👍

sstanisicTT:
I think Profiler is currently done quite nicely, however, I cant merge until I finish up porting Perf, which will take quite more work. Since this work doesn't impact the API in any visible way, I will be coming back to this later..

tests/python_tests/helpers/perf.py
```diff
@@ -18,127 +17,133 @@
...
 
+ALL_RUN_TYPES = [type for type in PerfRunType]
 
-@dataclass
-class PerfData:
-    unpack: PerfThreadData
-    math: PerfThreadData
-    pack: PerfThreadData
 
+def _stats_timings(perf_data: pd.DataFrame) -> pd.DataFrame:
```
fvranicTT: Would this work?
```python
def _stats_timings(perf_data: pd.DataFrame) -> pd.DataFrame:
    timings = perf_data.columns.drop("marker")
    result = perf_data.groupby("marker", as_index=False)[timings].agg(["mean", "std"])
    result.columns = ["marker"] + [f"{stat}({col})" 
                                    for col in timings 
                                    for stat in ["mean", "std"]]
    return result # no need to reset index thanks to `as_index = False` in groupy.
```

tests/python_tests/helpers/perf.py
```diff
@@ -18,127 +17,133 @@
...
+    return result.reset_index()
 
 
-def process_profiler_data(profiler_data) -> PerfData:
-    return PerfData(
-        unpack=_parse_thread(profiler_data.unpack),
-        math=_parse_thread(profiler_data.math),
-        pack=_parse_thread(profiler_data.pack),
-    )
+def _stats_l1_to_l1(data: ProfilerData) -> pd.Series:
```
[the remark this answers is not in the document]
sstanisicTT: I find the current solution much easier to understand, and i think we should keep it as is. The number of markers will probably be <5 and in total there probably won't be more than 150 entries in the table that is processed, so it's unlikely that this code will be the performance bottleneck.

If the Perf infra becomes a bottleneck, I'll optimize it further after profiling the code.

Issue #624, opened 2025-08-28
Transpose dest tests fail when Float data format tests run first, pass when Int32 tests run first


amokanTT:
While writing tests and sweeping llk_math_transpose_dest.h I have encountered the following issues:

---------------------------
[ISSUE PART 1]
[DIFFERENT TEST RESULTS DEPENDING ON TEST ORDER AND DATA TYPES]

When running input_format=output_format=Int32 tests first, and then float data format tests after, all of the tests pass successfully. When running the float data format tests first, some float data format tests fail.

The branch with the cpp and py transpose_dest tests is amokan/math_transpose_dest_sweep.
Issue #363 and the conversation there should provide context about the test cases covered in the transpose_dest test.

The tests that fail are:
1. Float32->Float32, dest_acc=Yes, math_transpose_faces=Yes/No, unpack_to_dest=Yes
2. Float32->Float16/Float16_b/Bfp8_b, dest_acc=Yes, math_transpose_faces=Yes/No, unpack_to_dest=Yes
These tests all have is_32bit=True when using the llk_math_transpose_dest API. They fail both on WH and BH.

The following tests fail as well:
1. Float16_b/Bfp8_b->Float16, dest_acc=No, math_transpose_faces=Yes, unpack_to_dest=No
These tests have dest_acc changed to Yes according to the data inference model, but is_32bit=false. Other than this, dest_acc=Yes is not tested with Float16/Float16_b/Bfp8_b. They fail only on WH.

-> See comment [DIFFERENT TEST RESULTS DEPENDING ON TEST ORDER AND DATA TYPES NO.2]

---------------------------
[ISSUE PART 2]
[LARGE INT32 VALUES]

Also, on both BH and WH, when the input format is Int32, the stimuli are generated to be values between 0 and 255. I tried testing out a higher number and in that case the Int32 tests fail as well. The exact scenario I tested was having the stimuli be a tensor filled with the value 1500000 = 0x0016e360, and the result I got was 0x0000e360, so half of the bytes were zeroed out. On the other hand, some Int32 tests with higher number pass, like 0xdead1234. 

-> See comment [LARGE INT32 VALUES NO.2]

amokanTT:
It seems like it is due to this line if (FlushDenormals && !(SrcBVal & 0xff)) SrcBVal = 0;
[...]

PR #674, opened 2025-09-19
zero flag fixes for transpose

sstanisicTT:
### Ticket

https://github.com/tenstorrent/tt-metal/issues/23553

### Problem description

For transpose to properly work we need to do a few things:
- Disable Zero Flag, preventing the hardware from flushing the mantissa if the exponent part of the 16b datum is 0
- https://github.com/tenstorrent/tt-isa-documentation/blob/019fa8560068f35f98e80368b0f473d72fa0c3b6/BlackholeA0/TensixTile/TensixCoprocessor/Dst.md?plain=1#L13

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/test_transpose_dest.py
+++ b/tests/python_tests/test_transpose_dest.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 # SPDX-License-Identifier: Apache-2.0
 
+from itertools import product
+
 import pytest
 import torch
 from helpers.device import (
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h
```diff
@@ -13,6 +13,7 @@
 #include "ckernel_template.h"
 #include "cmath_common.h"
 #include "llk_math_common.h"
+#include "llk_unpack_common_api.h"
```
nvelickovicTT: Where is this file? Is it a misspell?

amokanTT:
> Can you think of a test that targets these scenarios? AFAIK we don't have a test neither in this repo, nor in tt-metal. Could we add something like e.g. transpose dest test for FP32, like @lpremovicTT suggested?

cpp and py tests have been added to the PR. All of the tests with Float formats and a 32bit dest are skipped for now because they will fail due to the part 1 of issue #624 (chicken bit). The zero flag fix from this PR is tested on the following test case: Int32 and 32bit dest. This fix solved part 2 of issue #624.
Custom test stimuli is generated for this INT32 test case, because the standard INT32 stimuli range is narrow and does not exercise the problem well.

fvranicTT (approved):

PR #696, opened 2025-09-25
setup quasar infrastructure and enable running risc tests on quasar

lpremovicTT:
### What's changed
Modified test infra's build and run system as well as helpers to support running tests on quasar.
While at it, updated blackhole linker script to be in line with WH one and ones used by tt-exalens.
Directly added quasar hw specific files, since there is no public place to dynamically download them from like we do for WH and BH.
Improved register store and core reset handling in device.py.
Added a risc compute test that performs eltwise addition on three int 32 tiles using only risc cores (no tensix), this is meant as a quick infra smoke test (it runs on QSR as of this pr).
Minimal modifications of qsr_llks to make them build inside test infra environment.
Added assembly.yaml for QSR.

### Type of change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### Checklist


tests/Makefile
```diff
@@ -85,6 +90,15 @@ TO_UPPER = $(shell echo $(1) | tr '[:lower:]' '[:upper:]')
 # =========================
 .PHONY: all dis profiler clean always
 
+ifeq ($(ARCH),quasar)
```
fvranicTT, line 93: To avoid repetition, you could do something like:
```Makefile
COMMON_ELF_TARGETS := $(ELF_DIR)/unpack.elf \
                      $(ELF_DIR)/math.elf \
                      $(ELF_DIR)/pack.elf

COMMON_DIS_TARGETS := $(DIS_DIR)/unpack.S \
                      $(DIS_DIR)/math.S \
                      $(DIS_DIR)/pack.S

ifneq ($(ARCH),quasar)
COMMON_ELF_TARGETS += $(SHARED_ELF_DIR)/brisc.elf
COMMON_DIS_TARGETS += $(SHARED_DIS_DIR)/brisc.S
endif

all: $(COMMON_ELF_TARGETS)
dis: $(COMMON_DIS_TARGETS)
```

tests/helpers/include/perf.h
```diff
@@ -36,8 +36,13 @@ inline void _perf_unpack_loop_set_valid(uint32_t iterations)
     {
         constexpr uint32_t cond_clear_a = set_a ? ckernel::p_stall::SRCA_CLR : 0;
         constexpr uint32_t cond_clear_b = set_b ? ckernel::p_stall::SRCB_CLR : 0;
+#ifdef ARCH_QUASAR
+        TTI_SETDVALID((set_b << 1) | set_a);
```
fvranicTT, line 40: Lines 40 and 43 look the same. Could we then just guard the STALLWAIT?

tt_llk_quasar/common/inc/ckernel.h
```diff
@@ -270,9 +290,9 @@ inline void set_ttsync_enables(uint thread_id = 0xdeadface)
 
     if (thread_id > 3)
     {
-        FWLOG0(
-            "WARNING: automatically writing to all TTSync enable regs, because the given thread ID (or lack thereof) implies that you don't know what thread "
-            "you want");
+        // FWLOG0(
```
fvranicTT, line 293: Should we remove FWLOG* completely? We deleted it from BH/WH IIRC.

tests/python_tests/helpers/device.py
```diff
@@ -69,14 +64,82 @@ class BootMode(Enum):
...
+            case _:
+                raise ValueError(f"Invalid neo_id {neo_id} for Quasar architecture")
+    else:
+        if neo_id != 0:
+            raise ValueError(f"Invalid non zero neo_id for non Quasar architecture")
+        register_store = noc_block.get_register_store()
+    return register_store
+
+
+def get_soft_reset_mask(cores: list[RiscCore]):
```
fvranicTT, line 119: This could be written in a more Pythonic way:
```python
def get_soft_reset_mask(cores: list[RiscCore]):
    if INVALID_CORE in cores:
        raise ValueError(
            "Attempting to reset a core that doesn't exist on this chip"
        )
    return sum(1 << core.value for core in cores)
```

tests/sources/risc_compute_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#ifdef LLK_TRISC_MATH
+
+void run_kernel()
+{
+    int32_t* A = (int32_t*)buffer_A[1];
+    int32_t* B = (int32_t*)buffer_B[1];
+    int32_t* C = (int32_t*)buffer_Res[1];
+
+    for (int i = 0; i < 1024; i++)
```
fvranicTT, line 40: Not that it matters, but this could be a one-liner with std::transform 😛 
```cpp
std::transform(A, A + 1024, B, C, std::plus<int32_t>());
```

tt_llk_quasar/common/inc/ckernel.h
```diff
@@ -71,6 +75,22 @@ inline void semaphore_get(const uint8_t index)
     pc_buf_base[PC_BUF_SEMAPHORE_BASE + index] = 1;
 }
 
+// Register read (workaround for bug
+// tenstorrent/tensix#976
+// now handled by the compiler)
+// workaround is needed only for GS
```
fvranicTT, line 81: GS? Is this comment still valid?

tt_llk_quasar/common/inc/ckernel.h
sstanisicTT: I think we shouldn't remove it. The old ones were removed because we didn't know if they still made sense. The QSR ones are fresh. And they are going to be useful when we add support for logging on device.

tt_llk_quasar/common/inc/ckernel.h
lpremovicTT: I wasn't able to find the original bug this references, if its accurate these functions should be removed from all architectures. I ported them over to quasar for backward compatibility, along with the comment to remind me to remove them at some point.

fvranicTT: OK

fvranicTT (approved):

PR #738, opened 2025-10-15
Stall changing packer config till packer finishes & add uninit functions for unpack_tilizeA_B

amahmudTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/27504

### Problem description
The stallwait for waiting for packer to finish before changing config registers is used too far down the code, after some registers which affect packing are changed. May be this was an optimization of some sort where we knew which registers were latched and when. But according to some discussions in LLK meetings we decided to not use such optimizations. 

There is no unpack_tilzieA_B_uninit which is also needed for the relevant issue to be fixed.

### What's changed
Stall till packer finishes before changing relevant configuration  registers. 
Added unpack_tilize_A_B_uninit 

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
@@ -357,3 +357,26 @@ inline void _llk_unpack_tilizeA_B_(
         switch_config_context(unp_cfg_context);
     }
 }
+
+inline void _llk_unpack_tilizeA_B_uninit_(const std::uint32_t unpack_dst_format, const std::uint32_t face_r_dim = FACE_R_DIM)
+{
+    // Revert X dim value to default.
+    TT_SETADCXX(p_setadc::UNP_A, face_r_dim * FACE_C_DIM - 1, 0x0);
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
```diff
@@ -357,3 +357,26 @@ inline void _llk_unpack_tilizeA_B_(
         switch_config_context(unp_cfg_context);
     }
 }
+
+inline void _llk_unpack_tilizeA_B_uninit_(const std::uint32_t unpack_dst_format, const std::uint32_t face_r_dim = FACE_R_DIM)
```
amahmudTT: I think the y_stride value does not need to be touched even though it is initialized in unpack_tilize_A_B_init. It is not used anywhere except for unpack_tilize_A_B. If somebody wants to use it , it can be initialized specifically. In general it is left alone and probably contains zero or 1. It might be a problem if someone depends on the default value. But then again what is the default value ? It is not initialized at all.

PR #746, opened 2025-10-16
Fused mul and reduce #26434

vbabicTT:
### Fused mul and reduce #26434
https://github.com/tenstorrent/tt-metal/issues/26434

### Problem description
As part of https://github.com/tenstorrent/tt-metal/issues/25523, a fused eltwise multiply and sum-reduce operation is needed to optimize the depthwise convolution path.

Currently, the depthwise path in conv2d lacks a dedicated implementation and falls back to the conventional matmul-based approach. This PR introduces an optimized solution using element-wise multiplication followed by reduction, with both operations fused together for improved performance. By fusing these operations, we eliminate the unpacking overhead between stages, which is currently a major performance bottleneck. The operation processes three operands: activation, weight, and scalar. Based on initial performance estimates, this fused implementation could nearly double the throughput compared to the current proof-of-concept where multiplication and reduction are executed as separate operations.

### What's changed
This PR adds comprehensive support for fused binary eltwise and column reduction operations across multiple components:

**Eltwise Binary Math (`llk_math_eltwise_binary.h`)**: Implemented destination register reuse functionality, enabling seamless data flow from binary operations directly to reduction without intermediate steps(Packer/Unpacker).

**Reduce Math (`llk_math_reduce.h`)**: Added `_llk_math_reduce_column_()` method that performs column reduction directly on source operand A without requiring unpacker involvement. Also introduced `_llk_math_reduce_clear_dvalid_after_for_loop_()` for proper data valid flag management in fused operation contexts.

**Unpacker (`llk_unpack_AB.h`)**: Enhanced unpacker configuration to support the specialized data flow requirements of fused operations. No unpacking is actually done for the reduce operation, the unpacker simply sets the dvalid flags for the math to begin.

**SFPU Utilities (`ckernel_sfpu_fill.h`)**: Added destination initialization functions including `_populate_first_tile_with_zeroes_()` to properly prepare accumulation buffers for reduction. Note that while zero initialization can alternatively be achieved via destination flags (faster), the SFPU-based approach was implemented for enhanced debugging capabilities and explicit control during development.

**Implementation notes**: Several SETRWC (set window counter) instructions have been added throughout the implementation. While some may appear redundant, they proved invaluable for debugging destination addressing issues, a common source of complexity when implementing fused operations. Some of them can be deleted in future iterations. Also, when moving the scaler from destination register to operand B, the better thing to do would be to move only the very first face of the tile, since it is the only one being used as a scaler, since srcB counter is never incremented in the fused operation. The LLK changes are located in the `tt_llk_blackhole` directory and are implemented only for the Blackhole-specific architecture.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_fill.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_fill.h
@@ -50,10 +50,11 @@ inline void _calculate_fill_bitcast_(const uint32_t value_bit_mask)
     }
 }
 
-inline void _populate_first_tile_with_ones_() {
+inline void _populate_first_tile_with_ones_()
+{
     // Reset destination counter to ensure we're pointing to the first tile
     TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
-    
+
     // Use SFPU to fill with ones
     TT_SETC16(DEST_TARGET_REG_CFG_MATH_Offset_ADDR32, 0);
     _calculate_fill_<false, 32>(1.0f);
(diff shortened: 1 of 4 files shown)


tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_fill.h
```diff
@@ -49,4 +49,30 @@ inline void _calculate_fill_bitcast_(const uint32_t value_bit_mask)
         sfpi::dst_reg++;
     }
 }
+
+inline void _populate_first_tile_with_ones_()
```
fvranicTT: Can we parametrize this to avoid repetition?
```cpp
inline void _populate_first_tile_with_value_(float value) {
    TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
    TT_SETC16(DEST_TARGET_REG_CFG_MATH_Offset_ADDR32, 0);
    _calculate_fill_<false, 32>(value);
    TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
}
```

PR #797, opened 2025-11-03
Optimize Welfords LLK to hold intermediates for multiple groups at once

nsorabaTT:
### Ticket
Issue: https://github.com/tenstorrent/tt-metal/issues/29706
This change accompanies the metal PR https://github.com/tenstorrent/tt-metal/pull/31732
Please see the description of that PR as well for full context.

### Problem description
Current welford implementation updates the mean and variance as it scans an entire tile (or a subset of rows in the tile). When applied to groupnorm, there are instances when parts of a single tile belong to multiple groups. To save on data movement, it is prudent to calculate the online mean and var for each of these groups when we have the tile fetched.

### What's changed
This PR does the following:
1. Adds functions to maintain up to 16 mean and vars simultaneously in a tile. Earlier, the 32 mean values for a tile were stored in the first 4 rows of a tile in destination registers. With this change, the "rows" are picked according to `group_id`
2. Reduce welford algorithm's instructions from 9 to 8 by using MOV instructions. Use the entire replay buffer by directly operating out of LREG0, LREG1 etc. and placing these in the replay buffer.
3. Separate out load, compute and store to give better control to kernel writers and allow for optimizations.
4. Write down docstrings for all functions.
5. Use `TT_` and `TTI_` instructions and templates to move as many things as possible to compile time calculations.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_welfords.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_welfords.h
@@ -256,7 +256,7 @@ sfpi_inline void _calculate_welfords_block_w_offset_(
     constexpr uint32_t block_max_row_idx = block_min_row_idx + 4;
 
     // If the start_row and end_row don't intersect with this block, we don't need to process this.
-    [[unlikely]] if (((start_row >= block_max_row_idx) || (end_row <= block_min_row_idx)))
+    if (((start_row >= block_max_row_idx) || (end_row <= block_min_row_idx)))
     {
         return;
     }
@@ -365,7 +365,7 @@ template <std::size_t reciprocal_size>
 sfpi_inline void _calculate_welfords_partial_tile_(
     uint32_t start_idx, uint32_t start_row, uint32_t num_rows, const std::array<uint32_t, reciprocal_size>& reciprocal_lut)
 {
-    [[unlikely]] if (num_rows == 0)
+    if (num_rows == 0)
     {
         return;
     }
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_welfords.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_welfords.h
@@ -258,7 +258,7 @@ sfpi_inline void _calculate_welfords_block_w_offset_(
     constexpr uint32_t block_max_row_idx = block_min_row_idx + 4;
 
     // If the start_row and end_row don't intersect with this block, we don't need to process this.
-    [[unlikely]] if (((start_row >= block_max_row_idx) || (end_row <= block_min_row_idx)))
+    if (((start_row >= block_max_row_idx) || (end_row <= block_min_row_idx)))
     {
         return;
     }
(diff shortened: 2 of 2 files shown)


tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_welfords.h
```diff
@@ -44,300 +44,520 @@ struct FloatBits
...
+    // ---------------
+    // m2_{N+1} = m2_{N} + (x_{N+1} - mean_{N}) * (x_{N+1} - mean_{N+1})
+    // Let α = x_{N+1} - mean_{N} and β = x_{N+1} - mean_{N+1}
+    // Then m2_{N+1} = m2_{N} + α * β
 
-    // var_{N+1} = var_{N} + ...
-    //...(1/N+1) * (((x_{N+1} - mean_{N}) * (x_{N+1} - mean_{N+1})) - var_{N})
+    // 1. Calculate α = x_{N+1} - mean_{N}
+    // LREG4 = -1 * LREG4 + input_lreg
+    TTI_SFPMAD(ckernel::p_sfpu::LREG11 /*-1*/, ckernel::p_sfpu::LREG4, input_lreg, ckernel::p_sfpu::LREG4, 0);
```
fvranicTT: Why do we calculate alpha twice? First time you do it in line 141.

bbradelTT: Maybe add that to the comment.

fvranicTT: So, if I'm understanding it correctly, Welford's algorithm needs these values simultaneously:

1. mean_old (LREG4)
2. mean_new (needs storage)
3. alpha = x - mean_old (needs storage)
4. beta = x - mean_new (can overwrite input)
5. M2 (LREG5)
6. 1/(n+1) (LREG7)

With only LREG6 available as temporary storage, we're forced to:
1. Calculate alpha into LREG6
2. Overwrite it with mean_new
3. Recalculate alpha when needed for M2

fvranicTT:
> /codeowners ping

That doesn't work here. It's a workflow from metal.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_welfords.h
```diff
@@ -44,300 +44,518 @@ struct FloatBits
...
+ *                   processed in the tile. Should be 0 <= start_row <= 31.
+ * @param num_rows The number of rows to process. Should be 0 <= num_rows <= 32. Also,
+ *                 0 <= start_row + num_rows <= 32.
+ * @param reciprocal_lut The lookup table containing the reciprocals of the sample counts.
+ */
+template <std::size_t reciprocal_size>
+sfpi_inline void _calculate_welfords_partial_tile_(
+    uint32_t start_idx, uint32_t start_row, uint32_t num_rows, const std::array<uint32_t, reciprocal_size>& reciprocal_lut)
+{
+    [[unlikely]] if (num_rows == 0)
```
fvranicTT:
```suggestion
    if (num_rows == 0)
```
These are C++20 extensions, likely and unlikely. LLK should keep compatibility with C++17. We are not yet strictly enforcing this, but maybe we should.

nsorabaTT:
> I think wormhole requires nops in certain places, please double check if you got nops correcty for wormhole in every place it requires.

Yes, the wormhole version has NOPs after each instruction that required it, specifically `SFPMAD`. The blackhole version omits this.

This is why the same the welford computation is 8 instructions in wormhole but only 6 in blackhole.

I reviewed it once again.

fvranicTT (approved):

PR #803, opened 2025-11-05
[Test Infra] Quasar - unpack_unary_operand and pack tests, UNP_DEST safeguards

amokanTT:
### Problem description
Test llk_unpack_unary_operand and llk_pack API for Quasar. Make the use of UNP_DEST more intuitive and less error prone.

### What's changed
- Added unpack_unary_operand Quasar tests -> 192 tests (224, but 32 are skipped).
- Added pack Quasar tests -> 48 tests (56, but 8 are skipped)
- Added safeguards for UNP_DEST use.  -> Before, if unpacking to dest, UNP_DEST would have to be passed to the _llk_unpack_unary_operand_init_ function, and UNP_A would have to be passed to _llk_unpack_configure_unary_ and _llk_unpack_unary_operand_. Passing UNP_DEST to _llk_unpack_configure_unary_ would lead to incorrectly calling instructions related to UNPACKER2/srcS. Passing UNP_DEST to _llk_unpack_unary_operand_ would result in incorrectly setting PACKER0 as the targeted execution engine instead of UNPACKER0. Now, passing UNP_DEST is safe and results in intended behaviour.
- Added transpose_en configure in case of unpacking to dest. -> Before, setting the correct value for transpose in the configuration register for Unpacker0/1 engines was done only in case of UNP_A/UNP_B, but not UNP_DEST. When unpacking to dest, transpose is still not supported. Running a test with UNP_A and Transpose.Yes first and a test with UNP_DEST and Transpose.No second would result in a stale transpose configuration and produce incorrect results. Now, this unpacker config register is set for UNP_DEST as well.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/sources/quasar/pack_quasar_test.cpp
+++ b/tests/sources/quasar/pack_quasar_test.cpp
@@ -32,11 +32,7 @@ void run_kernel()
         set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_dvalid_client::FPU, dest_dvalid_client::PACK});
     }
 
-    buffer_descriptor_u bd_val;
-    for (uint i = 0; i < BD_NUM_WORDS; i++)
-    {
-        bd_val.words[i] = 0;
-    }
+    buffer_descriptor_u bd_val = {0};
 
     bd_val.f.l1_addr_16B = buffer_A[0] / 16;
     bd_val.f.format      = static_cast<uint8_t>(formats.unpack_src);
@@ -129,11 +125,7 @@ void run_kernel()
         set_up_dest_dvalid_per_thread<dest_dvalid_client::PACK>({dest_dvalid_client::FPU, dest_dvalid_client::PACK});
     }
 
-    buffer_descriptor_u bd_val;
-    for (uint i = 0; i < BD_NUM_WORDS; i++)
-    {
-        bd_val.words[i] = 0;
-    }
+    buffer_descriptor_u bd_val = {0};
     tdma_descriptor_t tdma_desc;
 
     bd_val.f.l1_addr_16B = buffer_Res[0] / 16;
--- a/tests/sources/quasar/unpack_unary_operand_quasar_test.cpp
+++ b/tests/sources/quasar/unpack_unary_operand_quasar_test.cpp
@@ -31,11 +31,7 @@ void run_kernel()
         set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_dvalid_client::FPU, dest_dvalid_client::PACK});
     }
 
-    buffer_descriptor_u bd_val;
-    for (uint i = 0; i < BD_NUM_WORDS; i++)
-    {
-        bd_val.words[i] = 0;
-    }
+    buffer_descriptor_u bd_val = {0};
 
     // Qsr has one transpose argument, if set it does both transpose faces and within face
     // The py test will set transpose faces and transpose within face to the same value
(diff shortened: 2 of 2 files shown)


tests/helpers/include/data_format_inference.h
```diff
@@ -12,8 +12,8 @@
 #include "build.h"
 #include "tensix_types.h"
 
-#if defined(ARCH_WORMHOLE) && defined(ARCH_BLACKHOLE)
-#error "Only one of ARCH_WORMHOLE or ARCH_BLACKHOLE can be defined"
+#if (defined(ARCH_WORMHOLE) && defined(ARCH_BLACKHOLE)) || (defined(ARCH_WORMHOLE) && defined(ARCH_QUASAR)) || (defined(ARCH_BLACKHOLE) && defined(ARCH_QUASAR))
```
fvranicTT: You can simplify this with one nice trick. `defined(X)` will return 1 if `X` is defined, 0 otherwise. You can simplify this long if using that.
```cpp
#if (defined(ARCH_WORMHOLE) + defined(ARCH_BLACKHOLE) + defined(ARCH_QUASAR)) > 1
#error "Only one of ARCH_WORMHOLE, ARCH_BLACKHOLE or ARCH_QUASAR can be defined"
#elif (defined(ARCH_WORMHOLE) + defined(ARCH_BLACKHOLE) + defined(ARCH_QUASAR)) == 0
#error "You must define either ARCH_WORMHOLE or ARCH_BLACKHOLE or ARCH_QUASAR"
#elif defined(ARCH_WORMHOLE)
constexpr bool is_blackhole = false;
constexpr bool is_wormhole  = true;
constexpr bool is_quasar    = false;
#elif defined(ARCH_BLACKHOLE)
constexpr bool is_blackhole = true;
constexpr bool is_wormhole  = false;
constexpr bool is_quasar    = false;
#else // ARCH_QUASAR
constexpr bool is_blackhole = false;
constexpr bool is_wormhole  = false;
constexpr bool is_quasar    = true;
#endif
```

tests/python_tests/helpers/llk_params.py
```diff
@@ -268,6 +268,21 @@ class EltwiseBinaryReuseDestType(Enum):
...
     No = "false"
     Yes = "true"
+
+
+class UnpackerEngine(Enum):
+    """
+    Enum for unpacker engine selection.
+    """
+
+    UNP_A = 0
```
fvranicTT: Let's try to stick to `PascalCase` we've been using so far (`EltwiseBinaryReuseDestType` should be fixed too, but in a separate PR).
So:
```python
class UnpackerEngine(Enum):
    """
    Enum for unpacker engine selection.
    """
    UnpA = 0
    UnpB = 1
    UnpS = 2
    UnpDest = 3
```

tt_llk_quasar/llk_lib/llk_unpack_common.h
```diff
@@ -19,22 +19,22 @@ inline void _llk_unpack_hw_configure_(const tdma_descriptor_t& tdma_desc_src)
 {
     static_assert(
         (UNP_SEL == p_unpacr::UNP_A) || (UNP_SEL == p_unpacr::UNP_B) || (UNP_SEL == p_unpacr::UNP_S) || (UNP_SEL == p_unpacr::UNP_DEST),
-        "UNP_SEL can only be set to p_unpacr::UNP_A/UNP_B/UNP_S");
+        "UNP_SEL can only be set to p_unpacr::UNP_A/UNP_B/UNP_S/UNP_DEST");
```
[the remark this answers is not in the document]
fvranicTT: Ported to the other testing env.

fvranicTT (approved):

PR #807, opened 2025-11-05
Add SFPU kernel for ADD top row of two tiles

skotaracTT:
### Ticket

https://github.com/tenstorrent/tt-metal/issues/29535#issuecomment-3479884917
### Problem description

The column reduction kernels store their results in the top row of each tile. If you want to add two reduced tiles together, you would be summing all the data elements in the tiles, even though only the top row matters. This is computationally heavy and inefficient.
### What's changed

Once the desired tiles have been reduced, if you want to accumulate their sums, this kernel computes the sum using only the first row of each tile. This approach greatly reduces the number of operations and cycles required.
### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/helpers/data_format_inference.py
+++ b/tests/python_tests/helpers/data_format_inference.py
@@ -255,9 +255,8 @@ def data_formats(
         num_iterations: The number of pipeline runs (iterations), determines list length
         unpacking_to_dest: Whether unpacking targets the destination register (default: False)
         chip_arch: The chip architecture (Wormhole or Blackhole). If None, will be detected automatically.
-        disable_format_inference: For majority of tests we trigger data format inference model, but for some tests we want to keep all formats the same.
-            When True we want to deactivate the data format inference and set all formats to be the same, no conversions will be performed.
-            This ensures that the input formats are the same ones that end up in dest for specific math kernels that support the formats we input.
+        disable_format_inference: When True, disables automatic data format inference and conversions, ensuring input formats are the same in dest.
+                                  Used for testing specific math kernels with explicit format requirements.
 
     Returns:
         A list of FormatConfig objects of length num_iterations


tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_add_top_row.h
```diff
@@ -0,0 +1,105 @@
...
+    TTI_SFPIADD(0, p_sfpu::LREG7, p_sfpu::LREG3, 4);
+
+    // Set up floating-point replay buffer (index 4, 8 instructions)
+    // Contains: 4 TT_SFPADD + 4 TTI_SFPNOP operations for adding top rows
+    lltt::record(4, 8);
+    TT_SFPADD(p_sfpu::LREG0, p_sfpu::LCONST_1, p_sfpu::LREG4, p_sfpu::LREG0, 0);
+    TTI_SFPNOP;
+    TT_SFPADD(p_sfpu::LREG1, p_sfpu::LCONST_1, p_sfpu::LREG5, p_sfpu::LREG1, 0);
+    TTI_SFPNOP;
+    TT_SFPADD(p_sfpu::LREG2, p_sfpu::LCONST_1, p_sfpu::LREG6, p_sfpu::LREG2, 0);
```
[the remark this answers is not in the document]
skotaracTT: No reason, should be TTI. Done

tests/python_tests/helpers/data_format_inference.py
```diff
@@ -240,6 +240,7 @@ def data_formats(
     num_iterations: int,
     unpacking_to_dest: bool = False,
     chip_arch: Optional[ChipArchitecture] = None,
+    trigger_format_inference: bool = True,
```
fvranicTT: Could you explain this parameter?

tests/python_tests/helpers/golden_generators.py
```diff
@@ -1151,6 +1152,11 @@ def __call__(
         t1 = to_tensor(operand1, data_format)
         t2 = to_tensor(operand2, data_format)
 
+        if operation == MathOperation.SfpuAddTopRow:
```
fvranicTT: Can we move this special handling into `_add_top_row()`?
That way we keep the `_call_()` code clean, as special cases should be handled separately within the op.

tests/python_tests/helpers/data_format_inference.py
skotaracTT: Yes I'll add a comment. For majority of tests we trigger data format inference model, but for some tests we want to keep all formats the same. In this case we want to deactivate the data format inference and set all formats to be the same. When the inference model was in header file there was a check for this however now we need it on the python end as well. For this specific case we want to ensure that the input formats are the same ones that end up in dest because AddTopRow kernel supports the formats we input.

tests/python_tests/helpers/data_format_inference.py
```diff
@@ -262,7 +263,7 @@ def data_formats(
     if chip_arch is None:
         chip_arch = get_chip_architecture()
 
-    if chip_arch == ChipArchitecture.QUASAR:
+    if chip_arch == ChipArchitecture.QUASAR or not trigger_format_inference:
```
fvranicTT: Isn't Quasar enough for now? Do we ever turn this on in case of Quasar?

skotaracTT: https://github.com/tenstorrent/tt-llk/pull/807#discussion_r2499849983

tests/python_tests/helpers/data_format_inference.py
```diff
@@ -254,6 +255,9 @@ def data_formats(
         num_iterations: The number of pipeline runs (iterations), determines list length
         unpacking_to_dest: Whether unpacking targets the destination register (default: False)
         chip_arch: The chip architecture (Wormhole or Blackhole). If None, will be detected automatically.
+        disable_format_inference: For majority of tests we trigger data format inference model, but for some tests we want to keep all formats the same.
```
fvranicTT:
```
disable_format_inference: When True, disables automatic data format inference and conversions, ensuring input formats are the same in dest. 
    Used for testing specific math kernels with explicit format requirements.
```
a bit more formal

fvranicTT (approved):

PR #809, opened 2025-11-05
perf :zap:: isolation in matmul benchmark

sstanisicTT:
### Problem description

Isolation modes for Matmul LLK benchmark.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -181,7 +181,7 @@ jobs:
       runs_on: ${{ matrix.runs_on }}
       test_splits: 4
       pytest_markers: "perf"
-      timeout_minutes: 120
+      timeout_minutes: 240
 
   check-all-green:
     name: "✅ Check all green"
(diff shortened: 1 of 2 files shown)


tests/helpers/include/perf.h
```diff
@@ -29,19 +30,39 @@ enum class PerfRunType
...
+    uint32_t wait_clear_a = set_a ? ckernel::p_stall::SRCA_CLR : 0;
+    uint32_t wait_clear_b = set_b ? ckernel::p_stall::SRCB_CLR : 0;
+
+#ifdef ARCH_QUASAR
+    TT_STALLWAIT(ckernel::p_stall::STALL_TDMA, wait_clear_a, wait_clear_b, 0);
+    TT_SETDVALID((set_b << 1) | set_a);
+
+#else
+    TT_STALLWAIT(ckernel::p_stall::STALL_TDMA, wait_clear_a | wait_clear_b);
+    TT_SETDVALID((set_b << 1) | set_a);
```
fvranicTT: This could go out of ifdef, right? But do we know that this actually works on quasar? Have you tried running it? If not, maybe we add Quasar support later?

sstanisicTT: I've added it do the ifdef because i felt it's easier to read, it could go out of the ifdef. I haven't tested any of these mocks on quasar yet, this is just code that I inherited from Lazar's fixes to this file

PR #819, opened 2025-11-07
Change PolynomialEvaluator to support vFloat coefficients

nmauriceTT:
### Ticket
##818

### Problem description

As mentioned in this [comment](https://github.com/tenstorrent/tt-metal/pull/31685#discussion_r2490797334), `PolynomialEvaluator::eval()` takes coefficients as array , which make it difficult to use with `sfpi::vFloat` (especially fixed and programmable constants). 

### What's changed

Replaced `PolynomialEvaluator` implementation with a recursive one that one variadic templates.

This allows for evaluations with mixed `float` and `sfpi::vFloat` coefficients such as the following:

```cpp
sfpi::vFloat result = PolynomialEvaluator::eval(
        val, 
        sfpi::vConst0, 
        0.999004364013671875, 
        3.0897438526153564453125e-2, 
        -0.4890659749507904052734375,
        sfpi::vConstFloatPrgm2,
        sfpi::vConstFloatPrgm1,
        sfpi::vConstFloatPrgm0);
```

Recursion depth should be okay: since C++11, the maximum recursion depth in templates is 1024 ([although gcc is limited to 900](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html)). 

Most polynomial approximations should not exceed 20 arguments (biggest one I've seen so far in metal is gelu with POLYVAL15). We should also be able to set an upper bound through `static_assert(sizeof...(Reset) < MAX_DEPTH)` (if `static_assert` evaluated before 'recursion').

### Type of change

- [ ] Documentation update

This is a technically a breaking change, but `PolynomialEvaluator` is not yet used in Metal (pending [accurate pow PR](https://github.com/tenstorrent/tt-metal/pull/30986) ). 

### Checklist

- [x] [All post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml) CI passes [OK](https://github.com/tenstorrent/tt-metal/actions/runs/19147554722)
- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml) CI passes (if applicable) [https://github.com/tenstorrent/tt-metal/actions/runs/19147559999]

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_polyval.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_polyval.h
@@ -47,7 +47,7 @@ struct PolynomialEvaluator
      * @note Coefficients can be either float, sfpi::vFloat, ... (scalar and sfpi typed arguments can be mixed)
      */
 
-    // Base case: f(x) = 0 (rank-0 polynomial)
+    // Base case: f(x) = 0 (empty polynomial)
     template <typename U>
     static constexpr auto eval(U x)
     {
@@ -57,7 +57,7 @@ struct PolynomialEvaluator
     template <typename U, typename Coefficient0>
     static constexpr auto eval(U x, Coefficient0 coeff0)
     {
-        // Base case: f(x) = coeff0 (rank-0 polynomial)
+        // Base case: f(x) = coeff0 (0-th degree polynomial)
         return coeff0;
     }
 
--- a/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_polyval.h
+++ b/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_polyval.h
@@ -47,7 +47,7 @@ struct PolynomialEvaluator
      * @note Coefficients can be either float, sfpi::vFloat, ... (scalar and sfpi typed arguments can be mixed)
      */
 
-    // Base case: f(x) = 0 (rank-0 polynomial)
+    // Base case: f(x) = 0 (empty polynomial)
     template <typename U>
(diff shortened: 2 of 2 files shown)


tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_polyval.h
```diff
@@ -23,46 +23,44 @@ constexpr T POLYVAL7(T coef6, T coef5, T coef4, T coef3, T coef2, T coef1, T coe
...
+     * @param rest Rest of the polynomial coefficients in ascending order of powers.
      * @return The value of the polynomial at the given point
      *
-     * @pre coef must point to at least N valid coefficients
-     * @pre N should be positive for meaningful results
-     *
+     * @note Coefficients can be either float, sfpi::vFloat, ... (scalar and sfpi typed arguments can be mixed)
      */
-    static constexpr T eval(const U* coefficients, T val)
+    template <typename U, typename First, typename... Rest>
```
fvranicTT: I like it. I have an upgrade to propose:
```cpp
// Base case
template <typename T, typename U>
static constexpr auto eval(T x, U first)
{
    return first;
}

// Recursive case: Horner's method
template <typename T, typename U, typename... Rest>
static constexpr auto eval(T x, U first, Rest... rest)
{
    return first + x * eval(x, rest...);
}
```
1. No if constexpr check at every recursion level
2. No need for sizeof...(Rest) == 0 check
3. The base case is explicit and separate

fvranicTT: We could even ask this in our interviews 😉

Issue #742, opened 2025-10-15
[Proposal] Revisiting how we parametrize tests


sstanisicTT:
As we implement full argument sweep testing, we face a problem: many input parameters constrain each other, leading to invalid combinations. Currently, we handle this by either adding many skip statements in tests or using a constraint solver - both approaches add complexity. 

We can simplify this by enhancing the @parametrize annotation to accept lambda functions that dynamically generate valid parameter values based on other parameters. Each lambda would receive the current values of its dependencies and return the valid options for those specific values (or an empty array if none exist). 

(The example is only meant to showcase how the parametrization would work, and is probably not correct code)

```python

# ... in some shared file ...

def get_valid_math_fidelities(format, operation):
    if operation in [MathOperation.Elwadd, MathOperation.Elwsub]:
        return [MathFidelity.LoFi]
    
    if format.input == DataFormat.Bfp8_b:
        return [MathFidelity.LoFi, MathFidelity.HiFi2]
    
    return [MathFidelity.LoFi, MathFidelity.HiFi2, MathFidelity.HiFi3, MathFidelity.HiFi4]


def get_valid_dest_accumulation_modes(format):
    outlier_inputs = [DataFormat.Bfp8_b, DataFormat.Float16_b]
    
    if (format.input in [DataFormat.Bfp8_b, DataFormat.Float16_b] and format.output == DataFormat.Float16):
        return [DestAccumulatio.On]

    return [DestAccumulation.Off, DestAccumulation.On]


# ... in test file ...

@parametrize(
    test_name="eltwise_binary_fpu_perf",
    formats=input_output_formats([DataFormat.Bfp8_b, DataFormat.Float16, DataFormat.Float16_b]),
    mathop=[MathOperation.Elwadd, MathOperation.Elwsub, MathOperation.Elwmul],
    math_fidelity=lambda formats, mathop: get_valid_math_fidelities(formats, mathop),
    dest_acc=lambda formats: get_valid_dest_accumulation_modes(formats),
)
def test_eltwise_binary_fpu(test_name, formats, mathop, math_fidelity, dest_acc):
    """Test element-wise binary operations with various configurations."""
    # Test implementation
    pass
```

This approach eliminates most skip statements without requiring a complex constraint solver for typical test cases. The only limitation is handling cyclic dependencies (where A depends on B and B depends on A), but these don't appear to be relevant for LLK tests in practice.

sstanisicTT:
[...]

PR #832, opened 2025-11-13
infra :construction:: lambda parametrization 

sstanisicTT:
### Ticket

#742 

### Problem description

Provides a way to specify constraints between input parameters of test using lamdas (example in linked issue)

This is backwards compatible (we never used functions as parameters to `@parametrize`)

I've added two simple constraints that I could think of.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/helpers/param_config.py
+++ b/tests/python_tests/helpers/param_config.py
@@ -3,7 +3,7 @@
 
 import inspect
 from itertools import product
-from typing import List, Optional, Tuple, TypedDict
+from typing import List, Tuple
 
 import pytest
 from helpers.chip_architecture import ChipArchitecture, get_chip_architecture
@@ -79,22 +79,6 @@ def format_combination_sweep(
         for pack_src in formats
         for pack_dst in formats
     ]
-
-
-class TestParamsConfig(TypedDict):
-    test_name: str
-    formats: Optional[List[FormatConfig]] = None
-    dest_acc: Optional[DestAccumulation] = None
-    approx_mode: Optional[List[str]] = None
-    mathop: Optional[List[str]] = None
-    math_fidelity: Optional[List[int]] = None
-    tile_count: Optional[int] = None
-    reduce_dim: Optional[List[str]] = None
-    pool_type: Optional[List[str]] = None
-    num_faces: Optional[List[int]] = None
-    dest_sync: Optional[DestSync] = None
-    tilize_en: Optional[Tilize] = None
-    dst_idx: Optional[List[int]] = None
 
 
 # When reading this file, keep in mind that parameter means FORMAL PARAMETER and argument means ACTUAL PARAMETER


tests/python_tests/helpers/param_config.py
```diff
@@ -96,51 +97,103 @@ class TestParamsConfig(TypedDict):
...
+
+
+def _param_dependencies(parameter: str, argument: any) -> set:
+    """Extract parameter names from a callable using introspection."""
+    if callable(argument):
+        return set(inspect.signature(argument).parameters.keys())
+    return set()
+
+
+def _topsort_dependencies(**kwargs: any) -> List[str]:
```
fvranicTT: We could improve this a bit with:
```python
def _topsort_dependencies(**kwargs: any) -> List[str]:
    # Build dependency graph: param -> set of params it depends on
    dependencies = {
        param: _param_dependencies(param, value) for param, value in kwargs.items()
    }

    topological = []
    resolved_set = set()

    while dependencies:
        resolved = [
            param for param, deps in dependencies.items() if not (deps - resolved_set)
        ]

        if not resolved:
            raise ValueError(
                f"Circular dependency detected among: {list(dependencies.keys())}"
            )

        # Add resolved parameters to topological order and resolved set
        resolved_set.update(resolved)
        topological.extend(resolved)

        # Remove resolved parameters from dependency graph
        for parameter in resolved:
            del dependencies[parameter]

    return topological
```
This way we switch from updating dependency sets to tracking a resolved_set. See how `if not deps` is changed to `if not (deps - resolved_set)` to check if dependencies are satisfied.

tests/python_tests/helpers/param_config.py
```diff
@@ -96,51 +97,103 @@ class TestParamsConfig(TypedDict):
...
-        ("multiple_tiles_eltwise_test", FormatConfig(DataFormat.Float16, DataFormat.Float16, DataFormat.Float16, DataFormat.Float16, DataFormat.Float16), DestAccumulation.Yes, ApproximationMode.Yes, None, None, None, None, None),
-        ("matmul_test", FormatConfig(DataFormat.Float16, DataFormat.Float16, DataFormat.Float16, DataFormat.Float16, DataFormat.Float16), DestAccumulation.Yes, ApproximationMode.No, None, None, None, None, None)
-    ]
+        List of tuples representing all valid parameter combinations
     """
+    topological = _topsort_dependencies(**kwargs)
 
-    format_combos = kwargs.get("formats", [])
-    dest_acc = kwargs.get("dest_acc", [])
+    def _resolve_param_values(param: str, resolved: dict) -> list:
```
fvranicTT: Instead of this, we could do:
```python
# Store parameter order upfront to avoid repeated kwargs.keys() calls
param_order = tuple(kwargs.keys())
topological = _topsort_dependencies(**kwargs)

# Pre-compute which parameters are callables and their dependencies
callable_params = {}
for param, value in kwargs.items():
    if callable(value):
        callable_params[param] = _param_dependencies(param, value)

# Create mapping from param name to index in param_order for fast lookup
param_to_idx = {param: idx for idx, param in enumerate(param_order)}
# Create mapping from topological index to param_order index
topo_to_order_idx = [param_to_idx[param] for param in topological]
```

And change `resolved` from `dict` to `list`:
```python
def _resolve_param_values(param: str, resolved_list: list) -> list:
    """Get possible values for a parameter given resolved dependencies."""
    value = kwargs[param]

    if param in callable_params:
        deps = callable_params[param]
        # Build constraint kwargs only for resolved dependencies
        constraint_kwargs = {}
        for dep in deps:
            dep_idx = param_to_idx.get(dep)
            if dep_idx is not None and resolved_list[dep_idx] is not None:
                constraint_kwargs[dep] = resolved_list[dep_idx]
        result = value(**constraint_kwargs)
        return result if isinstance(result, list) else [result]

    if isinstance(value, list):
        return value

    return [value]

def _solve_recursive(parameter_idx: int, resolved: list) -> List[Tuple]:
    """Recursively build combinations starting from parameter at idx."""
    if parameter_idx >= len(topological):
        # Return tuple of resolved parameters in the original order
        return [tuple(resolved)]

    # Get current parameter and its possible values
    parameter = topological[parameter_idx]
    param_idx_in_order = topo_to_order_idx[parameter_idx]
    arguments = _resolve_param_values(parameter, resolved)

    if not arguments:
        return []

    # For each possible value, recurse with next parameter
    combinations = []
    for argument in arguments:
        # Create new resolved list with updated value
        resolved_next = list(resolved)
        resolved_next[param_idx_in_order] = argument
        combinations.extend(_solve_recursive(parameter_idx + 1, resolved_next))

    return combinations

# Initialize resolved list with None values
initial_resolved = [None] * len(param_order)
return _solve_recursive(0, initial_resolved)
```

1. List copy is faster than dict copy (fewer allocations, no hash table rebuild)
2. Index-based access: resolved[idx] is O(1) direct memory access vs hash lookup
3. Pre-computed mappings: topo_to_order_idx avoids repeated lookups

tests/python_tests/helpers/param_config.py
```diff
@@ -96,51 +97,103 @@ class TestParamsConfig(TypedDict):
...
+            resolved_next[parameter] = argument
+            combinations.extend(_solve_recursive(parameter_idx + 1, resolved_next))
+
+        return combinations
+
+    return _solve_recursive(0, {})
 
 
 def parametrize(**kwargs: any):
     parameters = kwargs.keys()
```
fvranicTT:
```suggestion
    parameters = tuple(kwargs.keys())
```

tests/python_tests/helpers/constraints.py
```diff
@@ -0,0 +1,65 @@
...
+        So I'm not sure if they should also be handled here.
+    """
+
+    if (
+        formats.input in [DataFormat.Bfp8_b, DataFormat.Float16_b]
+        and formats.output == DataFormat.Float16
+    ):
+        return [DestAccumulation.Yes]
+
+    if formats.input in [DataFormat.Int32, DataFormat.UInt32]:
```
skotaracTT: Would it be possible to add a constraint that checks for 32bit input datums and sets unpack_to_dest=True as a parameter behind the scenes in case the user forgets to set this? Is this even possible to do, since we would need to pass it into test_config dictionary...should we just set this in test_config?

sstanisicTT: I didn't want to do this because technically we may want to test FP32 -> TF32 in source in which case unpack_to_dest=false IIRC.

skotaracTT: Great point! Three steps ahead 👍 User will know if they want to unpack to dest, no need for us to assume for or against in advance.

PR #842, opened 2025-11-17
refactor: replace z3 filtering to constraint-based approach

fvranicTT:
### Ticket
None

### Problem description
The `test_unpack_A.py` test suite was using Z3 solver to filter valid parameter combinations based on hardware constraints. This caused a significant performance bottleneck:
- ~11 minute delay before tests could start running
- z3 solver overhead for each parameter combination evaluation

Screenshot 2025-11-17 at 16 07 13
image

(left - old, right - with this change)

### What's changed
Replaced Z3-based constraint solving with standard Python boolean logic and optimized the filtering function for better performance. Filtering of valid combinations now completes in seconds.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -34,4 +34,3 @@ termcolor==3.1.0
 tomli==2.2.1
 torch==2.8.0
 typing-extensions==4.14.1
-z3-solver==4.13.4.0


vkrsmanovicTT:
Since we decided to remove Z3 constraint mechanisms due to complexity from the codebase, it can be removed from the: https://github.com/tenstorrent/tt-llk/blob/fvranicTT/simplify-constraint-solving/tests/requirements.txt

PR #856, opened 2025-11-19
Ldjurovic/sfpu reduce max patch

ldjurovicTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/26699

### Problem description
SFPU reduce max needed to be updated in order to be able to use subblock_w=1 to accumulate over whole column of tensor and only after that store results in dest.

### What's changed
Changed implementation of sfpu reduce column max LLK, and changed test case to resemble real SDPA use case. In future both will be expanded to be flexible and usable on random dimensions.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/test_sfpu_reduce_sdpa.py
+++ b/tests/python_tests/test_sfpu_reduce_sdpa.py
@@ -28,12 +28,9 @@ from helpers.utils import passed_test
     reduce_pool=[ReducePool.Max],  # Only MAX is supported for SDPA reduce
     input_dimensions=[
         [32, 32],
-        [32, 64],
         [64, 32],
-        [128, 64],
-        [64, 128],
-        [32, 128],
         [128, 32],
+        [256, 32],
     ],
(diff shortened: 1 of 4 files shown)


tests/python_tests/test_sfpu_reduce_sdpa.py
```diff
@@ -28,12 +28,9 @@
     reduce_pool=[ReducePool.Max],  # Only MAX is supported for SDPA reduce
     input_dimensions=[
         [32, 32],
-        [32, 64],
```
fvranicTT: Why are these being removed?

ldjurovicTT: Because algorithm changed and now is only for subblock_w = 1. I need this to be merged to unblock metal folks, and when we have some free time we can go back to this and modify it to work on random subblock_w. That shouldn't be too much work but isn't on our radar atm

PR #881, opened 2025-11-26
Create Sfpu Reduce Kernel API + Extend Max reduce kernel to support Int, UInt and Floats

skotaracTT:
### Ticket

[#29534](https://github.com/tenstorrent/tt-metal/issues/29534)
### Problem description

Max kernel implementation needed to support Max Op. Currently, the Max Op falls back to less optimized execution path because no specialized kernel exists in the backend. This limits performance, prevents proper integration in some pipelines, and blocks downstream features that rely on efficient reductions.
### What's changed

Incorporating correct and efficient Max reduction kernel (utilizing `LOADMACRO` for performance benefits) into Sfpu Reduce Kernels, so that the runtime can execute Max reductions natively, enabling full Op support and unblocking downstream model development.
- Extending Max kernel to support formats: Float32, Int32, UInt32, UInt16, BFloat16
- Created cohesive Sfpu Reduce LLK API for all supported reductions, incorporating DRY principles, clean dispatch. Making it more maintainable for new formats/pool_types and reducing bloat & repetition.

### Type of change

- [ ] Documentation update

### DIFF
--- a/tests/python_tests/conftest.py
+++ b/tests/python_tests/conftest.py
@@ -10,7 +10,6 @@ import pytest
 from helpers.chip_architecture import ChipArchitecture, get_chip_architecture
 from helpers.device import reset_mailboxes
 from helpers.format_config import InputOutputFormat
-from helpers.log_utils import _format_log
 from helpers.target_config import TestTargetConfig, initialize_test_target_from_pytest
 from ttexalens import tt_exalens_init
 from ttexalens.tt_exalens_lib import arc_msg
@@ -173,14 +172,6 @@ def pytest_sessionstart(session):
 
 
 def pytest_sessionfinish(session, exitstatus):
-    BOLD = "\033[1m"
-    YELLOW = "\033[93m"
-    RESET = "\033[0m"
-    if _format_log:
-        print(f"\n\n{BOLD}{YELLOW} Cases Where Dest Accumulation Turned On:{RESET}")
-        for input_fmt, output_fmt in _format_log:
-            print(f"{BOLD}{YELLOW}  {input_fmt} -> {output_fmt}{RESET}")
-
     test_target = TestTargetConfig()
     if not test_target.run_simulator:
         # Send ARC message for GO IDLE signal. This should decrease device clock speed.
--- a/tests/python_tests/helpers/constraints.py
+++ b/tests/python_tests/helpers/constraints.py
@@ -1,0 +1,65 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from helpers.format_config import DataFormat
+from helpers.llk_params import DestAccumulation, MathFidelity, MathOperation
+
+
+def get_valid_dest_accumulation_modes(formats):
+    """
+    Base constraints for Dest Accumulation modes.
+
+    - Dest accumulation must be ENABLED for the following format combinations:
+        - Input format has B type exponent (bfp8_b, float16_b)
+        - Output format is A type exponent (float16)
+        Reason: HW limitation, Packer cannot convert expB to expA, so we convert it to Float32 first as intermediate. (Source???)
+    - Dest accumulation must be ENABLED for the following format combination:
+        - Input format is 32bit integer format (int32, uint32)
+        Reason: HW limitation, Unpacker cannot unpack 32bit integer formats into SrcA and SrcB registers
+    - Otherwise it can be ENABLED or DISABLED
+
+    NOTE: There are more combos that fit this rule, but aren't handled in the codebase
+        So I'm not sure if they should also be handled here.
+    """
+
+    if (
+        formats.input in [DataFormat.Bfp8_b, DataFormat.Float16_b]
+        and formats.output == DataFormat.Float16
+    ):
+        return [DestAccumulation.Yes]
+
+    if formats.input in [DataFormat.Int32, DataFormat.UInt32]:
+        return [DestAccumulation.Yes]
+
+    return [DestAccumulation.No, DestAccumulation.Yes]
+
+
+def get_valid_math_fidelities(format, operation, PERF_RUN: bool = False):
+    """
+    Base constraints for Math Fidelity modes.
+
+    - Regular mode:
+        - Math fidelity must be LoFi for ElwAdd and ElwSub operations
+        - Otherwise it can be LoFi, HiFi2, HiFi3, HiFi4.
+
+    - Performance mode:
+        - Ignores Math fidelity settings that are higher than necessary for full precision
+    """
+
+    if operation in [MathOperation.Elwadd, MathOperation.Elwsub]:
+        return [MathFidelity.LoFi]
+
+    # HiFi2 will multiply BFP8 and BFP8_b in full precision, skip HiFi3 and HiFi4
+    if PERF_RUN and format.input in [DataFormat.Bfp8_b, DataFormat.Bfp8]:
+        return [MathFidelity.LoFi, MathFidelity.HiFi2]
+
+    # todo: once support for any of these is added, add them here
+    # LoFi will multiply FP8, BFP4, BFP4a, BFP2, BFP2a in full precision, skip HiFi2 and higher
+
+    return [
+        MathFidelity.LoFi,
+        MathFidelity.HiFi2,
+        MathFidelity.HiFi3,
+        MathFidelity.HiFi4,
+    ]
--- a/tests/python_tests/helpers/log_utils.py
+++ b/tests/python_tests/helpers/log_utils.py
@@ -1,9 +1,0 @@
-# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
-# SPDX-License-Identifier: Apache-2.0
-
-_format_log = []
-
-
-def add_to_format_log(input_fmt, output_fmt):
-    global _format_log
-    _format_log.append((input_fmt, output_fmt))
--- a/tests/python_tests/helpers/param_config.py
+++ b/tests/python_tests/helpers/param_config.py
@@ -1,12 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
+import inspect
 from itertools import product
-from typing import List, Optional, Tuple, TypedDict
+from typing import List, Tuple
 
 import pytest
 from helpers.chip_architecture import ChipArchitecture, get_chip_architecture
-from helpers.log_utils import add_to_format_log
+from typing_extensions import deprecated
 
 from .format_config import (
     DataFormat,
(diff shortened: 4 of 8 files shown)


tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -255,252 +209,353 @@ inline void sfpu_reduce_max_col_configure_addrmod(uint32_t num_cols)
...
-            calculate_reduce_int<pool_type, reduce_dim, InstrModLoadStore::INT32>();
+            sum_columns<6>();                               // Integer replay buffer
+            TT_SFPIADD(0, p_sfpu::LREG4, p_sfpu::LREG0, 4); // LREG0 = upper_face_sums + lower_face_sums (int)
         }
-        else if constexpr (format == DataFormat::UInt32)
+        else
         {
-            calculate_reduce_int<pool_type, reduce_dim, InstrModLoadStore::INT32_2S_COMP>();
+            sum_columns<6>();                                                            // Float replay buffer (no NOPs needed for Blackhole)
+            TT_SFPADD(p_sfpu::LREG0, p_sfpu::LCONST_1, p_sfpu::LREG4, p_sfpu::LREG0, 0); // LREG0 = upper + lower (float)
```
fvranicTT:
```suggestion
            TTI_SFPADD(p_sfpu::LREG0, p_sfpu::LCONST_1, p_sfpu::LREG4, p_sfpu::LREG0, 0); // LREG0 = upper + lower (float)
```

tests/python_tests/test_sfpu_reduce.py
```diff
@@ -62,48 +65,54 @@ def test_sfpu_reduce(
...
             f"Skipping negative_numbers=True for unsigned format {formats.input_format}"
         )
 
     input_dimensions = dimension_combinations
     torch_format = format_dict[formats.input_format]
 
-    src_A, src_B, tile_cnt = generate_stimuli(
-        formats.input_format, formats.input_format, input_dimensions=input_dimensions
+    # STIMULI GENERATION
+    tile_cnt = input_dimensions[0] * input_dimensions[1] // 1024
```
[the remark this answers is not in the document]
skotaracTT: Agreed

tests/python_tests/test_sfpu_reduce.py
```diff
@@ -62,48 +65,54 @@ def test_sfpu_reduce(
...
+    dst_dim = input_dimensions
+    if reduce_pool != ReducePool.Max:
+        dst_dim = [32, tile_cnt * 32]
+    src_A = untilize_block(src_A, formats.input_format, dst_dim)
+    src_A = tilize_block(
+        src_A, dst_dim, stimuli_format=formats.input_format
+    ).flatten()  # Input tensor is tilized in dst register
+    src_A_untilized = untilize_block(
+        src_A, formats.input_format, dst_dim
+    )  # Passed into golden since PyTorch library has no concept of tilization
```
fvranicTT:
```suggestion
    src_A_untilized = untilize_block(src_A, formats.input_format, input_dimensions)
```
Why can't we just untilize once and skip the tilize/flatten/untilize cycle?

tests/python_tests/test_sfpu_reduce.py
```diff
@@ -116,7 +125,7 @@ def test_sfpu_reduce(
         "unpack_to_dest": True,
         "tile_cnt": tile_cnt,
         "disable_format_inference": True,
-        "add_top_row": add_top_row,
+        "unpack_to_dest": True,
```
fvranicTT: We won't have a test for add_top_row function anymore? Or you plan to add that in a separate PR?

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -255,252 +209,353 @@ inline void sfpu_reduce_max_col_configure_addrmod(uint32_t num_cols)
...
-    // SFPU LOADMACRO CONFIGURATION
+/**
+ * @brief Initialization for SFPU reduce MAX kernel.
+ *        Sets up LOADMACRO sequences for compare-and-swap operations, configures address modifiers,
+ *        and records replay buffers for efficient column-wise maximum reduction.
+ *
+ * @tparam INSTRUCTION_MODE The instruction mode for integer and float formats: INT32, INT32_2S_COMP, LO16, FP32, FP16B
+ * @param num_cols The number of columns to process (typically 32 for a single tile, or multiple of 32 for block operations)
+ */
+template <InstrModLoadStore INSTRUCTION_MODE, PoolType pool_type>
```
fvranicTT: Since `pool_type` isn't used, we could remove it completely.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -4,234 +4,188 @@
...
+    {
+        return InstrModLoadStore::INT32_2S_COMP;
+    }
+    else if constexpr (format == DataFormat::UInt16)
+    {
+        return InstrModLoadStore::LO16; // UInt16 uses LO16 mode for unsigned 16-bit
+    }
+    else
+    {
+        return InstrModLoadStore::INT32; // Default fallback
```
fvranicTT: How about static_assert-ing here instead of returning a default? Or even better, use `is_supported_reduce_format()` at the beginning to fail at compile-time for unsupported formats?

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -4,234 +4,188 @@
...
-        // Final result stored in top row of upper face (first row in dest) - no intermediate storage needed
+constexpr uint NUM_FACES                   = 4;
+constexpr uint UPPER_FACE_ADDRS[NUM_FACES] = {0, 0, 16, 16};   // Face 0, 0, 1, 1
+constexpr uint LOWER_FACE_ADDRS[NUM_FACES] = {32, 32, 48, 48}; // Face 2, 2, 3, 3
+constexpr uint COLUMN_OFFSETS[NUM_FACES]   = {0, 2, 0, 2};     // even, odd, even, odd
 
-        const uint UPPER_FACE_ADDR = UPPER_FACE_ADDRS[i];
-        const uint LOWER_FACE_ADDR = LOWER_FACE_ADDRS[i];
-        const uint COLUMN_OFFSET   = COLUMN_OFFSETS[i];
+constexpr uint TILE_SIZE     = 32;
```
fvranicTT: `TILE_SIZE` is defined, but I don't see it used anywhere.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -255,252 +209,353 @@ inline void sfpu_reduce_max_col_configure_addrmod(uint32_t num_cols)
...
+
+        // Upper face column summation (LREG0-3) with float operations
+        TTI_SFPADD(p_sfpu::LREG2, p_sfpu::LCONST_1, p_sfpu::LREG3, p_sfpu::LREG2, 0); // LREG2 = (LREG2 * 1) + LREG3
+        TTI_SFPADD(p_sfpu::LREG1, p_sfpu::LCONST_1, p_sfpu::LREG2, p_sfpu::LREG1, 0); // LREG1 = (LREG1 * 1) + LREG2
+        TTI_SFPADD(p_sfpu::LREG0, p_sfpu::LCONST_1, p_sfpu::LREG1, p_sfpu::LREG0, 0); // LREG0 = (LREG0 * 1) + LREG1
+
+        // Lower face column summation (LREG4-7) with float operations
+        TTI_SFPADD(p_sfpu::LREG6, p_sfpu::LCONST_1, p_sfpu::LREG7, p_sfpu::LREG6, 0); // LREG6 = (LREG6 * 1) + LREG7
+        TTI_SFPADD(p_sfpu::LREG5, p_sfpu::LCONST_1, p_sfpu::LREG6, p_sfpu::LREG5, 0); // LREG5 = (LREG5 * 1) + LREG6
+        TTI_SFPADD(p_sfpu::LREG4, p_sfpu::LCONST_1, p_sfpu::LREG5, p_sfpu::LREG4, 0); // LREG4 = (LREG4 * 1) + LREG5
```
fvranicTT: We might benefit from interleaving adds from upper and lower face. That way we don't have read-after-write dependencies.

```cpp
// Interleaved pattern - independent ops between dependent ones
TTI_SFPADD(LREG2, LCONST_1, LREG3, LREG2, 0);  // upper: LREG2 = LREG2 + LREG3
TTI_SFPADD(LREG6, LCONST_1, LREG7, LREG6, 0);  // lower: LREG6 = LREG6 + LREG7
TTI_SFPADD(LREG1, LCONST_1, LREG2, LREG1, 0);  // upper: LREG1 = LREG1 + LREG2

TTI_SFPADD(LREG5, LCONST_1, LREG6, LREG5, 0);  // lower: LREG5 = LREG5 + LREG6
TTI_SFPADD(LREG0, LCONST_1, LREG1, LREG0, 0);  // upper: LREG0 = LREG0 + LREG1
TTI_SFPADD(LREG4, LCONST_1, LREG5, LREG4, 0);  // lower: LREG4 = LREG4 + LREG5
```

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -4,234 +4,188 @@
...
+        TTI_SFPMOV(0, p_sfpu::LREG0, p_sfpu::LREG1, 0);                                      // Save original value for sign check
+        TTI_SFPABS(0, p_sfpu::LREG0, p_sfpu::LREG0, 0);                                      // Get absolute value of LREG0
+        TTI_SFPSHFT(-AVG_SHIFT_AMOUNT & AVG_SHIFT_MASK, p_sfpu::LREG0, p_sfpu::LREG0, 0b01); // Perform logical right shift by 5 bits (divide by 32)
+
+        // Restore sign if original value was negative
+        // Check if original value was negative (sign bit set)
+        TTI_SFPSETCC(0, p_sfpu::LREG1, 0, 4);               // Set condition code if original sign bit is 0 (positive)
+        TTI_SFPCOMPC(0, 0, 0, 0);                           // Invert condition code (now true if original was negative)
+        TTI_SFPIADD(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 6); // Negate LREG0 if condition is true
+        TTI_SFPENCC(0, 0, 0, 0);                            // Clear condition codes
```
fvranicTT: The `SFPSETCC` reads `LREG1`, which was written by `SFPMOV`. Meanwhile, `SFPABS` and `SFPSHFT` only operate on `LREG0`. This gives us the opportunity to interleave ops:

```cpp
TTI_SFPMOV(0, p_sfpu::LREG0, p_sfpu::LREG1, 0);
TTI_SFPABS(0, p_sfpu::LREG0, p_sfpu::LREG0, 0);
TTI_SFPSETCC(0, p_sfpu::LREG1, 0, 4);
TTI_SFPSHFT(-AVG_SHIFT_AMOUNT & AVG_SHIFT_MASK, p_sfpu::LREG0, p_sfpu::LREG0, 0b01);
TTI_SFPCOMPC(0, 0, 0, 0);
TTI_SFPIADD(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 6);
TTI_SFPENCC(0, 0, 0, 0);
```
WDYT?

tests/python_tests/test_sfpu_reduce.py
skotaracTT: Yup, I had this here when I was debugging so I could visualize tilized inputs better. Will remove it 👍

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
skotaracTT: You're right, I was using it for an implementation that I now optimized. Will remove

skotaracTT: Removed 👍

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
skotaracTT: I remember we talked about these kinds of solutions in order to reduce the need for NOPs! Great, I'll give it a try on blackhole and wormhole without the NOPs 👌

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -54,16 +54,16 @@ template <InstrModLoadStore INSTRUCTION_MODE>
 inline void load_face_data(uint upper_face_addr, uint lower_face_addr, uint column_offset)
 {
     // Load upper face data (Face 0 or Face 1) into LREG0-3
-    TT_SFPLOAD(p_sfpu::LREG0, INSTRUCTION_MODE, ADDR_MOD_7, upper_face_addr + column_offset);                     // rows 0-3
-    TT_SFPLOAD(p_sfpu::LREG1, INSTRUCTION_MODE, ADDR_MOD_7, upper_face_addr + column_offset + ROWS_PER_LOAD);     // rows 4-7
-    TT_SFPLOAD(p_sfpu::LREG2, INSTRUCTION_MODE, ADDR_MOD_7, upper_face_addr + column_offset + 2 * ROWS_PER_LOAD); // rows 8-11
-    TT_SFPLOAD(p_sfpu::LREG3, INSTRUCTION_MODE, ADDR_MOD_7, upper_face_addr + column_offset + 3 * ROWS_PER_LOAD); // rows 12-15
+    TTI_SFPLOAD(p_sfpu::LREG0, INSTRUCTION_MODE, ADDR_MOD_7, upper_face_addr + column_offset);                     // rows 0-3
```
fvranicTT: These must stay TT, as they contain runtime parameters.

tests/python_tests/test_sfpu_reduce.py
```diff
@@ -62,48 +65,54 @@ def test_sfpu_reduce(
...
         )
 
     input_dimensions = dimension_combinations
     torch_format = format_dict[formats.input_format]
 
-    src_A, src_B, tile_cnt = generate_stimuli(
-        formats.input_format, formats.input_format, input_dimensions=input_dimensions
+    # STIMULI GENERATION
+    tile_cnt = input_dimensions[0] * input_dimensions[1] // 1024
+    max, min = 127, (
```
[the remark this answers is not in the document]
skotaracTT: Extended to [-1000,1000]
In a sum kernel for one tile the max we can have is 32,000 which is just below upper bound  for UInt16 👍

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -255,252 +209,353 @@ inline void sfpu_reduce_max_col_configure_addrmod(uint32_t num_cols)
...
-    TTI_SFPLOADMACRO(5, InstrModLoadStore::FP16B, ADDR_MOD_7, 2);
-
-    TTI_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::FP16B, ADDR_MOD_7, 16);
-    TTI_SFPLOAD(p_sfpu::LREG1, InstrModLoadStore::FP16B, ADDR_MOD_7, 18);
+    TTI_INCRWC(0, 4, 0, 0);
+    TTI_SFPLOADMACRO(5, INSTRUCTION_MODE, ADDR_MOD_7, 2);
+    TTI_SFPLOAD(p_sfpu::LREG0, INSTRUCTION_MODE, ADDR_MOD_7, 16);
+    TTI_SFPLOAD(p_sfpu::LREG1, INSTRUCTION_MODE, ADDR_MOD_7, 18);
     TTI_SFPSWAP(0 /*unused*/, p_sfpu::LREG7 /*lreg_src_c*/, p_sfpu::LREG1 /*lreg_dest*/, 1 /*instr_mod1*/);
     TTI_SFPNOP;
```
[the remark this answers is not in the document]
skotaracTT: Heavy test failure without the NOPs. We tried though. 
In SFPU Spec for blackhole it says the following
```
The following are some notable limitations of the SFPU: 
Some SFPU instructions take 2 cycles to execute  
 - SFPSWAP and some flavors of SFPSHFT2 always require 2 cycles, and should always be padded with a NOP 
```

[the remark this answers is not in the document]
nvelickovicTT: This is surprising to me. I know that SWAP and SHFT2 take 2 cycles, but also BH has logic that inserts the delay itself, without programmer interefering.
If we have to do it on our own, then there is a part we don't fully understand.
That being said, I have 2 questions here:

1. Are these issued as part of LOADMACRO? If so, perhaps the logic doesn't work for that pipeline.
2. When you experiment and remove NOPs, do you also reflect on the size of recorded instruction buffer in replay?

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
nvelickovicTT: That is weird, we should figure out why this happens. What Filip proposed should work according to what we know.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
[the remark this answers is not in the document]
skotaracTT: Yes I agree with you, the only explanation I have is that BH Sfpu spec explicitly states this for SWAP and SHFT2 so this might be a specific case. We should have a sit down and investigate this further. 
To address your questions:

1. These specific SWAPS are not configured for LOADMACRO, they get executed separately after the LOADMACRO.
2. When removing NOPs the size of the buffer reflects this, you need to reduce the number of instructions.

tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
```diff
@@ -255,252 +207,353 @@ inline void sfpu_reduce_max_col_configure_addrmod(uint32_t num_cols)
...
+        TTI_SFPIADD(0, p_sfpu::LREG1, p_sfpu::LREG0, 4); // LREG0 = LREG0 + LREG1
+        TTI_SFPIADD(0, p_sfpu::LREG5, p_sfpu::LREG4, 4); // LREG4 = LREG4 + LREG5
+    }
+    else
+    {
+        // Float replay buffer: 6 instructions (no NOPs needed for Blackhole)
+        lltt::record(0, 6);
+
+        // Upper face column summation (LREG0-3) with float operations
+        TTI_SFPADD(p_sfpu::LREG2, p_sfpu::LCONST_1, p_sfpu::LREG3, p_sfpu::LREG2, 0); // LREG2 = (LREG2 * 1) + LREG3
```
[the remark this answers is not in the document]
skotaracTT: Almost, passes on BH but fails for UInt16 and Float16_b on WH. I can make an if-else case for these specific formats I know that Ops team isn't concerned with Uint16 and float16_b as much as Int32. But then again, there is no test in compute kernel API to see if this change would reflect well in that layer.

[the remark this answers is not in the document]
skotaracTT: I added this to blackhole 👍

PR #901, opened 2025-12-01
Added packer test with sweeped parameters.

uvelimirovicTT:
### What's changed
Packer kernel did not have a sufficient amount of tests. 
This PR configures a packer test (not untilizing) that is supposed to represent a starting point for overall explicit packer testing.
Test will be extended little by little.

### Type of change
Added test and a new class (PackGolden) in golden_generators.

### DIFF
--- a/tests/sources/pack_test.cpp
+++ b/tests/sources/pack_test.cpp
@@ -28,7 +28,7 @@ void run_kernel()
     for (int i = 0; i < TILE_CNT; ++i)
     {
         _llk_unpack_A_<BroadcastType::NONE, false, EltwiseBinaryReuseDestType::NONE, unpack_to_dest>(
-            L1_ADDRESS(buffer_A[i]), 0, formats.unpack_src, formats.unpack_dst);
+            L1_ADDRESS(buffer_A[i]), formats.unpack_src, formats.unpack_dst);
     }
 }
 
@@ -52,9 +52,9 @@ void run_kernel()
 {
 // copy srca to dest
 #ifdef ARCH_BLACKHOLE
-    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, false, is_int_fpu_en>(0, 0, num_faces, formats.math);
+    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, false, is_int_fpu_en>(num_faces, formats.math);
 #else
-    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, is_int_fpu_en>(0, 0, num_faces, formats.math);
+    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, is_int_fpu_en>(num_faces, formats.math);
 #endif
     _llk_math_pack_sync_init_<DST_SYNC, is_fp32_dest_acc_en>();
     _llk_math_hw_configure_<false, false>(formats.math, formats.math);


tests/python_tests/helpers/golden_generators.py
```diff
@@ -882,6 +882,54 @@ def __call__(
...
 
+# This is the same as DataCopyGolden but with ReLU applied. It's made as a separate class
+# so future pack testing can extend it as required without affecting (unnecessarily bloating) DataCopyGolden.
+@register_golden
+class PackGolden:
+    def __call__(
+        self,
+        operand1,
+        data_format,
+        num_faces: int = 4,
```
fvranicTT: Please add validation for num_faces:

```python
if num_faces not in [1, 2, 4]:
    raise ValueError(f"num_faces must be 1, 2, or 4, got {num_faces}")
```

tests/sources/pack_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+#ifdef ARCH_BLACKHOLE
+    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, false, is_int_fpu_en>(0, 0, num_faces, formats.math);
+#else
+    _llk_math_eltwise_unary_datacopy_init_<DataCopyType::A2D, is_fp32_dest_acc_en, BroadcastType::NONE, is_int_fpu_en>(0, 0, num_faces, formats.math);
+#endif
+    _llk_math_pack_sync_init_<DST_SYNC, is_fp32_dest_acc_en>();
+    _llk_math_hw_configure_<false, false>(formats.math, formats.math);
+    for (int i = 0; i < TILE_CNT; ++i)
+    {
+        _llk_math_wait_for_dest_available_<DST_SYNC>();
```
amokanTT: We usually put `_llk_math_wait_for_dest_available_` and `_llk_math_dest_section_done_` outside of the TILE_CNT loop. We write LLK tests so that they do not exceed the number of tiles that fit into dest. So math should be able to write an x amount of tiles into dest and have pack pack out the x amount of tiles after. In this test, math writes one tile to dest, pack reads one tile from dest and so on, which is less performant. 
Try changing this, if everything else is fine, this should work.

tests/sources/pack_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    _llk_pack_dest_init_<DST_SYNC, is_fp32_dest_acc_en, DstTileFaceLayout::RowMajor>();
+#else
+    _llk_pack_hw_configure_<is_fp32_dest_acc_en, false>(formats.pack_src, formats.pack_dst, 16 * 16 * 4, FACE_R_DIM, num_faces, false, false, RELU_CONFIG);
+    _llk_pack_init_<false, false, DstTileFaceLayout::RowMajor, false>(formats.pack_dst, FACE_R_DIM, num_faces);
+    _llk_pack_dest_init_<DST_SYNC, is_fp32_dest_acc_en, DstTileFaceLayout::RowMajor, false>();
+#endif
+
+    for (int i = 0; i < TILE_CNT; ++i)
+    {
+        _llk_packer_wait_for_math_done_();
```
amokanTT: Same here, try putting `_llk_packer_wait_for_math_done_` and `_llk_pack_dest_section_done_` outside of the TILE_CNT loop.

tests/sources/pack_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+        _llk_pack_<DST_SYNC, is_fp32_dest_acc_en, false>(DST_INDEX, L1_ADDRESS(buffer_Res[i]));
```
amokanTT: DST_INDEX+i

tests/python_tests/helpers/golden_generators.py
```diff
@@ -882,6 +882,59 @@ def __call__(
...
+            raise ValueError(f"num_faces must be 1, 2, or 4, got {num_faces}")
+
+        torch_format = format_dict[data_format]
+
+        height, width = input_dimensions[0], input_dimensions[1]
+
+        # Handle partial faces (face_r_dim < 16) as single tiles
+        if face_r_dim < 16:
+            tile_cnt = 1
+            tile_size = height * width
```
amokanTT: To be honest, I am not sure if this will work or need additional changes in the future. If we are not yet testing face_r_dim < 16 then maybe we should only merge what is tested and definitely works.

amokanTT (approved):

Issue #958, opened 2025-12-17
[Perf] Add eltwise unary sfpu perf tests


ndivnicTT:
There is a gap between functional and perf tests which use SFPU. The idea is to bridge the gap and measure math ops performances over SFPU by adding perf variant of tests over SFPU. This issue is used for eltwise unary SFPU tests.

Issue #657, opened 2025-09-16
Cleanup: Standardise tiny tiles


rtawfik01:
Tiny tiles has a few common function arguments that are not standardized. For example:

```
template <bool zero_srcA = false>
inline void _llk_unpack_tilizeA_B_(
    std::uint32_t unpA_src_format,
    std::uint32_t face_r_dim,
    std::uint32_t narrow_tile,
    std::uint32_t base_address_a,
    std::uint32_t address_b,
    std::uint32_t tile_index_a,
    std::uint32_t tile_index_b,
    std::uint32_t block_ct_dim,
    std::uint32_t num_faces = 4)
```
or
```
inline void _llk_unpack_AB_matmul_hw_configure_(
    const std::uint32_t unpA_src_format,
    const std::uint32_t unpB_src_format,
    const std::uint32_t unpA_dst_format,
    const std::uint32_t unpB_dst_format,
    const std::uint32_t unpA_face_r_dim             = FACE_R_DIM,
    const std::uint32_t unpB_face_r_dim             = FACE_R_DIM,
    const std::uint32_t within_face_16x16_transpose = 0,
    const std::uint32_t unpA_num_faces              = 4,
    const std::uint32_t unpB_num_faces              = 4,
    const std::uint32_t unpA_tile_size              = 0,
    const std::uint32_t unpB_tile_size              = 0)
```

The convention was usually the following:

```
num_faces = number of faces in a tile, range can be 1-4
face_r_dim = number of rows in a face, values = [1,2,4,8,16]
narrow_tile = set this flag to true if the number of columns < 32, example for a tile of 32x16
partial_face = set this flag to true if number of rows in face < 16
tile_size = number of datums in a tile (example 32x32 tile = 1024)
```

Now that quasar kernels are in this repo, and quasar has a standard struct:
```
struct TileShape
{
    uint32_t num_faces;
    uint32_t face_r_dim;
    uint32_t face_c_dim;
    bool narrow_tile;
};
```

It is better if this system is transferred to WH/BH, because it cleans up the function arguments, and it makes a standard for passing down tile information to all LLK APIs.

@nvelickovicTT @fvranicTT @skotaracTT fyi, this is a cleanup item that can be set as bounty?

vmilicevicTT:
To be revisited after unified LLK APIs are done

PR #959, opened 2025-12-17
[Perf] SFPU - eltwise unary tests

ndivnicTT:
### Ticket
#958

### Problem description
There is a gap between functional and perf tests which use SFPU. The idea is to bridge the gap and measure math ops performances over SFPU by adding perf variant of tests over SFPU. This issue is used for eltwise unary SFPU tests.

### What's changed
Added perf test for SFPU eltwise unary ops.

### Type of change

- [*] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### Checklist

### DIFF
--- a/.github/workflows/trigger-tt-metal-bump.yml
+++ b/.github/workflows/trigger-tt-metal-bump.yml
@@ -15,6 +15,7 @@ jobs:
   trigger-ci:
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    if: ${{ !startsWith(github.event.head_commit.message, '[skip uplift]') }}
     steps:
       - name: Send dispatch event to tt-metal
         uses: actions/github-script@v7
--- a/tests/setup_testing_env.sh
+++ b/tests/setup_testing_env.sh
@@ -49,14 +49,14 @@ download_headers() {
     echo "Downloading headers for ${chip_arch}..."
     mkdir -p "$header_dir/internal"
 
-    local base_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/tt-1xx/${chip_arch}"
+    local base_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/internal/tt-1xx/${chip_arch}"
     local headers=("cfg_defines.h" "dev_mem_map.h" "tensix.h" "tensix_types.h")
 
-    local risc_attribs_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/risc_attribs.h"
+    local risc_attribs_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/internal/risc_attribs.h"
 
     local specific_url=""
     if [[ "$chip_arch" == "wormhole" ]]; then
-        specific_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/tt-1xx/${chip_arch}/wormhole_b0_defines"
+        specific_url="https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/internal/tt-1xx/${chip_arch}/wormhole_b0_defines"
     fi
 
     for header in "${headers[@]}"; do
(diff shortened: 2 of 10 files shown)


tests/python_tests/helpers/stimuli_generator.py
```diff
@@ -120,6 +120,35 @@ def generate_face_matmul_data(
     return src
 
 
+def calculate_tile_and_face_counts(
```
skrsmanovicTT: I understand this is just clean up and separating this into function so no real work required here.
More as reminder that we should check tiny tile support in general in our test infra. For example common use case is to have face_r_dim < 16 but also to have more than 1 tile. For example 1, 2, 4, 8 x 64(or more) input tensor..

Reem provided good overview of tiny tile support in LLK #657 also new effort is planned for future regrading tile size refactoring, documented here: https://docs.google.com/document/d/1OB4ZGOBVtuvMytRqSIHOetS_qS8rA2by45_5SPyMVoQ/edit?usp=sharing

tests/python_tests/perf_eltwise_unary_sfpu.py
```diff
@@ -0,0 +1,165 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""
```
skrsmanovicTT: This was not the way we did it so far, so test should not have commented out variants if they are not relevant. If they are relevent then we should test them. 
@sstanisicTT What do you think about this ?

Maybe we could have different types of perf sweeps, as we talked earlier L1_to_L1 only for CI gating, full sweep for nightly run, sweep over tile_loop for getting dependency etc... But commenting some options in and out is not cleanest solution

tests/python_tests/perf_eltwise_unary_sfpu.py
```diff
@@ -0,0 +1,165 @@
...
+    loop_factor=[
+        # 1,
+        # 2,
+        # 4,
+        # 8,
+        16,
+        # 32,
+        # 64,
+        # 128,
+    ],  # Number of iterations to run the test in order to minimize measurement noise
```
skrsmanovicTT: nitpick: I would not call it noise since deviation of our measurements is close to 0, this is more due to error introduced with time measurement methodology we have(or let's say due to time measurement precision). Since every timestamp reading introduce overhead of ~30 cycles if our zone of interest is short we can get imprecise results. @sstanisicTT is this stated correctly ?

tests/python_tests/perf_sfpu_reduce_sdpa.py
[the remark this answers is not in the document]
ndivnicTT: Lazar and I had a sync offline, we agreed to drop sdpa related changes from this PR. Lazar will follow up with refactoring of this test. I created an issue which describes it: https://github.com/tenstorrent/tt-llk/issues/964

tests/python_tests/helpers/perf.py
```diff
@@ -27,9 +27,14 @@ class PerfRunType(Enum):
...
     PACK_ISOLATE = 4
     L1_CONGESTION = 5
+    L1_CONGESTION_MATH_SYNC = 6
 
 
 ALL_RUN_TYPES = [type for type in PerfRunType]
+CONGESTIONS_RUN_TYPES = [
+    PerfRunType.L1_CONGESTION,
+    PerfRunType.L1_CONGESTION_MATH_SYNC,
+]
```
sstanisicTT: Is this used anywhere, I can't seem to find it?

tests/python_tests/perf_eltwise_unary_sfpu.py
sstanisicTT: For now I would enable them all (and increase the CI workflow runtime probably), and we will pick later which types should be executed on PR and which are executed in NIGHTLY, especially cause this code is probably a much smaller offender than `perf_matmul`.

tests/python_tests/perf_eltwise_unary_sfpu.py
sstanisicTT: Something like  "to minimize profiler overhead in measurement" would be better.

tests/python_tests/perf_eltwise_unary_sfpu.py
```diff
@@ -0,0 +1,165 @@
...
+    face_r_dim=[FACE_R_DIM],
+    num_faces=[NUM_FACES],
+    input_dimensions=[
+        # [32, 32],  # tile_cnt: 1
+        # [64, 32],  # tile_cnt: 2
+        # [64, 64],  # tile_cnt: 4
+        [128, 64],  # tile_cnt: 8
```
sstanisicTT: I'd enable the smaller tile counts just so that we can track how higher tile_cnt improves bandwidth, and remove tile_cnt > 8 for now until we get faster CI

PR #971, opened 2025-12-22
Added relu sweep to pack_test.

uvelimirovicTT:
### Problem description
 - So far, packer ReLu activation function hasn't had a proper test. 

### What's changed
 - Added ReLu parameter sweep to test_pack.py with a helper function.
 - Also, to be able to test ReLu properly, we need matrices with negative values. Therefore negative value support has been added to golden_generators.

### DIFF
--- a/tests/python_tests/test_pack.py
+++ b/tests/python_tests/test_pack.py
@@ -38,8 +38,8 @@ def is_relu_threshold_tolerance_issue(
     result_tensor,
     relu_config,
     intermediate_format,
-    rtol=0.05,
-    atol=0.05,
+    rtol=0.01,
+    atol=0.01,
 ):
     """
     Check if test failure is due to threshold rounding/format conversion issues in ReLU.
@@ -56,8 +56,8 @@ def is_relu_threshold_tolerance_issue(
         golden_tensor: Expected output tensor
         result_tensor: Actual hardware output tensor
         relu_config: The ReLU configuration value
-        rtol: Relative tolerance for threshold proximity checks (default 0.05)
-        atol: Absolute tolerance for threshold proximity checks (default 0.05)
+        rtol: Relative tolerance for threshold proximity checks (default 0.01)
+        atol: Absolute tolerance for threshold proximity checks (default 0.01)
     Returns:
         bool: True if all mismatches are near-threshold rounding issues, False otherwise
     """


tests/python_tests/helpers/stimuli_generator.py
```diff
@@ -50,15 +55,17 @@ def generate_random_face(
             else:
                 # random for both faces
                 srcA_face = torch.rand(size, dtype=format_dict[stimuli_format])
+                if negative_values:
```
fvranicTT: Can you explain a bit what does this do?

tests/python_tests/test_pack.py
```diff
@@ -30,6 +33,82 @@
...
+        result_is_zero = result_tensor[mismatches] == 0.0
+        acceptable = (golden_is_zero & result_near_threshold) | (
+            result_is_zero & golden_near_threshold
+        )
+    else:  # For MAX_THRESHOLD_RELU: Check if mismatch is near the threshold clamp point
+        both_near_threshold = golden_near_threshold & result_near_threshold
+        allowed_difference = torch.isclose(
+            golden_tensor[mismatches], result_tensor[mismatches], rtol=rtol, atol=atol
+        )
+        acceptable = both_near_threshold & allowed_difference
```
fvranicTT: Wouldn't:
```python
acceptable = golden_near_threshold & result_near_threshold
```
In else branch be enough?
Here is why:

1. The mismatches mask is created as:
```python
mismatches = ~torch.isclose(golden_tensor, result_tensor, rtol=rtol, atol=atol)
```
This means mismatches is `True` at positions where the golden and result tensors are not close to each other (using rtol=0.05, atol=0.05).

2. Later, the code does:
```python
allowed_difference = torch.isclose(golden_tensor[mismatches], result_tensor[mismatches], rtol=rtol, atol=atol)
```
This is checking if the same pairs of values that already failed the torch.isclose test would now pass the torch.isclose test using the exact same tolerance parameters.
Finally, `allowed_difference` will always be False (or an all-False tensor) because you're re-checking the same condition with the same parameters that already failed 😉

tests/python_tests/test_pack.py
```diff
@@ -30,6 +33,71 @@
 from helpers.utils import passed_test
 
 
+def is_relu_threshold_tolerance_issue(
```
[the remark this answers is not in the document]
uvelimirovicTT: You are right to be suspicious about it. 

One way to do this without this function would be to completely simulate data transformation and rounding that the hardware does in the packer pipeline when applying ReLu.

Current implementation leads us to have around 2-3 errors every 100 tests or so. It's a rare occurrence, but unfortunately exists which requires some solution to the issue. 

I agree that this isn't the best way to solve this issue but it works just fine.

Perhaps posting this as an issue might be okay so in the future we could implement a precise software simulation of packer that takes into account all data transformation.

PR #1042, opened 2025-12-31
fix. Nightly fixes + coverage report generation script

ajankovicTT:
### Problem description
- test_unpack_matmul.py was failing and thus the whole nightly test suite.
- It wasn't easy to tests just a *.cpp file without passsing any arguments to constructors of `TestConfig` and `ProfilerConfig`.
- There wasn't a simple way to see coverage data for one are few tests variants

### What's changed
- Nightly is now passing
- You can write something like `configuration = ProfilerConfig("sources/profiler_overhead_test.cpp")` and your test would execute as intended.
- You can execute `./generate_coverage_report.sh` to produce lcov html report of your coverage test run.

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/python_tests/helpers/profiler.py
+++ b/tests/python_tests/helpers/profiler.py
@@ -490,9 +490,9 @@ class ProfilerConfig(TestConfig):
         self,
         test_name: str,
         formats: FormatConfig = None,
-        run_types: list[PerfRunType] = None,
-        templates: set[TemplateParameter] = None,
-        runtimes: set[RuntimeParameter] = None,
+        run_types: list[PerfRunType] = [],
+        templates: list[TemplateParameter] = [],
+        runtimes: list[RuntimeParameter] = [],
         variant_stimuli: StimuliConfig = None,
         unpack_to_dest=False,
         disable_format_inference=False,
@@ -511,15 +511,6 @@ class ProfilerConfig(TestConfig):
             disable_format_inference,
             dest_acc,
         )
-
-        if runtimes is None:
-            runtimes = []
-
-        if templates is None:
-            templates = []
-
-        if run_types is None:
-            run_types = []
 
         self.passed_templates = templates
         self.passed_runtimes = runtimes
--- a/tests/python_tests/helpers/test_config.py
+++ b/tests/python_tests/helpers/test_config.py
@@ -134,7 +134,7 @@ class TestConfig:
     _BUILD_DIRS_CREATED: ClassVar[bool] = False
 
     # When the infrastructure itself needs to be tested, some functionality like compiling the artefacts and writing them
-    # to tmpfs can be skipped. This flag is used to ship such code to enable fast execution of infra tests.
+    # to tmpfs can be skipped (eg. object, elf and coverage data files etc.). This flag is used to skip such code to enable fast execution of infra tests.
     INFRA_TESTING: ClassVar[bool] = False
 
     # === Addresses ===
@@ -288,8 +288,8 @@ class TestConfig:
         self,
         test_name: str,
         formats: FormatConfig = None,
-        templates: set[TemplateParameter] = None,
-        runtimes: set[RuntimeParameter] = None,
+        templates: list[TemplateParameter] = [],
+        runtimes: list[RuntimeParameter] = [],
         variant_stimuli: StimuliConfig = None,
         boot_mode: BootMode = BootMode.DEFAULT,
         profiler_build: ProfilerBuild = ProfilerBuild.No,
@@ -303,7 +303,7 @@ class TestConfig:
         )
 
         if test_name is None:
-            raise ValueError(
+            raise RuntimeError(
                 "test_name argument needs to be passed in order to resolve which C++ file is compiled"
             )
 
@@ -339,18 +339,17 @@ class TestConfig:
 
         self.runtime_format = "@"
 
-        if self.runtimes is not None:
-            for parameter in self.runtimes:
-                field_str, param_field_types = parameter.convert_to_struct_fields()
-                lines.append(field_str)
-                self.runtime_format += param_field_types
+        for parameter in self.runtimes:
+            field_str, param_field_types = parameter.convert_to_struct_fields()
+            lines.append(field_str)
+            self.runtime_format += param_field_types
 
         lines.append("};")
 
         self.runtime_params_struct = lines
 
     def write_runtimes_to_L1(self, location: str = "0,0"):
-        if self.runtimes is None:
+        if len(self.runtimes) == 0:
             return
 
         argument_data = []
@@ -696,7 +695,7 @@ class TestConfig:
             unpack_size_a = TILE_SIZES.get(self.formats.input_format, 128)
             unpack_size_b = TILE_SIZES.get(self.formats.input_format, 128)
 
-        if self.runtimes is not None:
+        if len(self.runtimes) > 0:
             itd_param = next(
                 (param for param in self.runtimes if isinstance(param, IN_TILE_DIMS)),
                 None,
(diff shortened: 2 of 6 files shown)


tests/python_tests/conftest.py
```diff
@@ -198,13 +208,46 @@ def pytest_runtest_makereport(item, call):
...
 
+                stack_trace = []
+                for entry in call.excinfo.traceback:
+                    path_str = str(entry.path)
+                    # Skip pytest and pluggy internal frames
+                    if "_pytest" in path_str or "pluggy" in path_str:
+                        continue
+
+                    try:
+                        file_path = entry.path.relative_to(Path.cwd())
```
fvranicTT: You can hoist out cwd from the loop.

tests/python_tests/helpers/profiler.py
```diff
@@ -512,18 +512,24 @@ def __init__(
...
-
         self.passed_templates = templates
         self.passed_runtimes = runtimes
         self.current_run_type = None
-        self.run_configs = [
-            (templates.copy() + [PERF_RUN_TYPE(run_type)], runtimes.copy(), run_type)
-            for run_type in run_types
-        ]
+
+        if run_types is not None:
```
fvranicTT: You can test this with a set, no need to run in loop (O(n) vs O(n*m) for loop):
```python
if run_types is not None:
    unsupported = set(run_types) - ProfilerConfig.SUPPORTED_RUNS
    assert not unsupported, (
        f"ERROR: run_types {unsupported} not implemented. "
        f"Supported: {set(ProfilerConfig.SUPPORTED_RUNS)}"
    )
    .
    .
    .
```
You also need to worry about templates.copy() and runtimes.copy() now that the two could be None which would cause a runtime error.
Why do we need this change?

tests/python_tests/helpers/test_config.py
```diff
@@ -696,11 +722,16 @@ def generate_build_header(self) -> str:
         return "\n".join(header_content)
 
     def build_elfs(self):
-        self.build_shared_artefacts()
 
         VARIANT_DIR = TestConfig.ARTEFACTS_DIR / self.test_name / self.variant_id
+        header_content = self.generate_build_header()
         done_marker = VARIANT_DIR / ".build_complete"
 
+        if TestConfig.INFRA_TESTING:
```
fvranicTT: What will this easter egg do? It's OK to test something once, but it could leave a backdoor in the code to skip some steps.

tests/python_tests/test_unpack_matmul.py
```diff
@@ -190,22 +190,13 @@ def test_unpack_matmul(math_fidelity, matmul_config, workers_tensix_coordinates)
             tile_count_res=matmul_config.tile_dimensions.tile_cnt,
         ),
         dest_acc=dest_acc,
-        unpack_to_dest=formats.input_format.is_32_bit(),
```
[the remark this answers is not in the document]
ajankovicTT: This is the fix. I didn't properly refactor the test the first time.

tests/requirements.txt
```diff
@@ -6,6 +6,7 @@ filelock==3.20.1
 fsspec==2025.7.0
 iniconfig==2.1.0
 jinja2==3.1.6
+junitparser==4.0.2
```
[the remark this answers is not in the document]
ajankovicTT: It's used for merged junit xml reports. Before we had `pip install junitparser` in the workflow, I moved it to requirements.txt to have it be more neat.

tests/python_tests/helpers/test_config.py
```diff
@@ -132,8 +132,10 @@ class TestConfig:
     MODE: ClassVar[TestMode] = TestMode.DEFAULT
     SKIP_JUST_FOR_COMPILE_MARKER: ClassVar[str] = "SKIPPED_JUST_FOR_COMPILE"
     _BUILD_DIRS_CREATED: ClassVar[bool] = False
+    INFRA_TESTING: ClassVar[bool] = False
```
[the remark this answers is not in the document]
ajankovicTT: I explained in the easter egg comment what it's for. I'll put a comment

tests/requirements.txt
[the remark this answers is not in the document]
fvranicTT: And we still do, if you are referring to `setup-and-test.yml`. Please remove it from there then 😉

tests/python_tests/test_sfpu_reduce.py
```diff
@@ -76,6 +77,11 @@ def test_sfpu_reduce(
     dimension_combinations,
     workers_tensix_coordinates,
 ):
+
+    # https://github.com/tenstorrent/tt-llk/issues/1040
+    if reduce_pool in [ReducePool.Average, ReducePool.Min] and TestConfig.WITH_COVERAGE:
+        pytest.skip()
```
fvranicTT:
```suggestion
        pytest.skip(reason="https://github.com/tenstorrent/tt-llk/issues/1040")
```

ajankovicTT: I like this

tests/python_tests/test_profiler_primitives.py
```diff
@@ -31,13 +30,10 @@ def assert_marker(
...
 def test_profiler_primitives(workers_tensix_coordinates):
 
-    configuration = ProfilerConfig(
-        "sources/profiler_primitives_test.cpp",
-        input_output_formats([DataFormat.Float16])[0],
-        variant_stimuli=StimuliConfig(
-            [], DataFormat.Float16, [], DataFormat.Float16, DataFormat.Float16, 1, 1
-        ),
-    )
+    if TestConfig.MODE == TestMode.PRODUCE:
```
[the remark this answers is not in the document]
fvranicTT: Why was it executed in both stages? Is that the case for both profiler tests and why does it apply only to them?

tests/generate_coverage_report.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+genhtml /tmp/tt-llk-build/merged_coverage.info --output-directory ../../coverage_report
```
fvranicTT: What will happen if the file doesn't exist? Could we please check that before calling genhtml?

tests/python_tests/conftest.py
```diff
@@ -198,13 +208,46 @@ def pytest_runtest_makereport(item, call):
...
 
+                stack_trace = []
+                for entry in call.excinfo.traceback:
+                    path_str = str(entry.path)
+                    # Skip pytest and pluggy internal frames
+                    if "_pytest" in path_str or "pluggy" in path_str:
+                        continue
+
+                    try:
+                        file_path = entry.path.relative_to(Path)
```
fvranicTT:
```suggestion
                        file_path = entry.path.relative_to(Path.cwd())
```
Would be correct. And we would still be calling it in every loop iteration. Can it be hoisted out?

ajankovicTT: It makes very usefull prints in the error messages that are not so verbose. Will fix

tests/python_tests/conftest.py
fvranicTT:
```suggestion
                stack_trace = []
                current_working_dir = Path.cwd()
                for entry in call.excinfo.traceback:
                    path_str = str(entry.path)
                    # Skip pytest and pluggy internal frames
                    if "_pytest" in path_str or "pluggy" in path_str:
                        continue

                    try:
                        file_path = entry.path.relative_to(current_working_dir)
```

tests/python_tests/helpers/test_config.py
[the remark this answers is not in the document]
fvranicTT: If that is the only thing that's skipped, I'd rather call it skip_building_elfs.

tests/python_tests/test_profiler_primitives.py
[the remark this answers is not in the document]
fvranicTT: Can you add a comment in the code explaining why we have this skip? Just having a skip doesn't mean much to someone reading the code.

PR #1094, opened 2026-01-14
Added transpose of faces for srcA when BType is COL or ROW.

uvelimirovicTT:
### Problem description
 - When running sweep test for unpack_AB api, there is a problem on blackhole when we turn on transpose and we have COL and ROW broadcast for srcB.
 - On Wormhole the transpose in srcA happens with no issue, while on blackhole the option for srcA transpose with those two broadcasts on srcB was simply left out for some reason.
 - This test discovered that hole:
 https://github.com/tenstorrent/tt-llk/pull/1103

### What's changed
_llk_unpack_AB_mop_config_ for blackhole has a new feature in 
    if constexpr (BType == BroadcastType::COL)
and
    else if constexpr (BType == BroadcastType::ROW)
branches

### Important:
This change is needed for the above test to pass. Otherwise if blackhole shouldn't support transpose by default for those two if branches then test should skip that scenario.

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_unpack_AB.h
+++ b/tt_llk_blackhole/llk_lib/llk_unpack_AB.h
@@ -22,24 +22,19 @@ template <BroadcastType BType = BroadcastType::NONE>
 inline void _llk_unpack_AB_mop_config_(const bool transpose_of_faces = false, const std::uint32_t num_faces = 4, const bool narrow_tile = false)
 {
     LLK_ASSERT(num_faces == 1 || num_faces == 2 || num_faces == 4, "num_faces must be 1, 2, or 4");
-<<<<<<< HEAD
-    static constexpr std::uint32_t unpack_srca           = TT_OP_UNPACR(SrcA, 0b1, 0, 0, 0, 1, 1, p_unpacr::RAREFYB_DISABLE, 0, 0, 0, 0, 1);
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/llk_lib/llk_unpack_AB.h
```diff
@@ -21,24 +21,38 @@ template <BroadcastType BType = BroadcastType::NONE>
...
-    if constexpr (BType == BroadcastType::COL)
+    if (transpose_of_faces)
     {
-        static constexpr uint unpack_srcb_set_z = TT_OP_SETADCZW(0b010, 0, 0, 0, 2, 0b0001);
-        const uint32_t outerloop                = num_faces < 4 ? 1 : 2;
-        const uint32_t innerloop                = num_faces < 2 ? 1 : 2;
-        ckernel_template tmp(outerloop, innerloop, unpack_srca);
-        tmp.set_start_op(unpack_srcb);
-        if (narrow_tile)
+        LLK_ASSERT(num_faces == 4, "num_faces must be 4 when transpose_of_faces is true");
```
ndivnicTT: You can move this to line 80 and delete if (transpose_of_faces)

uvelimirovicTT: I spoke with @ldjurovicTT. We concluded it's the best if we leave it to assert as early as possible.

PR #1121, opened 2026-01-19
Expand pack test to support face_r_dim < 16 and num_faces = 1 and num_faces = 2.

uvelimirovicTT:
### Problem description
 - Pack test always relied on default num_faces=4 and default face_r_dim=16 parameters.
 - This PR expands the test to support sweeped num_faces and face_r_dim parameters.

### What's changed
 - Expanded test_pack.py
 - Changed INPUT_DIMENSIONS class in test_input_variants.py to support different row and column numbers. They were fixed to 32,32.

### Noticed
 - Our whole testing infrastructure supports only single-tiled input matrices when having face_r_dim < 16.
 - This should be changed in the future.

### DIFF
--- a/tests/python_tests/helpers/golden_generators.py
+++ b/tests/python_tests/helpers/golden_generators.py
@@ -1054,7 +1054,10 @@ class PackGolden:
 
         height, width = input_dimensions[0], input_dimensions[1]
 
-        tile_cnt = (height // 32) * (width // 32)
+        tile_width = 32
+        tile_height = 32 if face_r_dim == 16 else face_r_dim
+
+        tile_cnt = (height // tile_height) * (width // tile_width)
         tile_size = height * width // tile_cnt
 
         # Calculate elements based on variable face dimensions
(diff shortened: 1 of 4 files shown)


tests/python_tests/helpers/golden_generators.py
```diff
@@ -1054,7 +1054,10 @@ def __call__(
 
         height, width = input_dimensions[0], input_dimensions[1]
 
-        tile_cnt = (height // 32) * (width // 32)
+        tile_width = 32
+        tile_height = 32 if face_r_dim == 16 else face_r_dim
+
+        tile_cnt = (height // tile_height) * (width // tile_width)
```
rtawfik01: Can we have these not as hardcoded constants? I believe you have another PR that standardises these

tests/python_tests/test_pack.py
```diff
@@ -120,30 +120,52 @@ def is_relu_threshold_tolerance_issue(
...
     dest_index,
     workers_tensix_coordinates,
 ):
 
+    if face_r_dim < 16 and (
+        formats.input_format == DataFormat.Bfp8_b
+        or formats.output_format == DataFormat.Bfp8_b
+    ):
+        pytest.skip(
+            "Bfp8_b format is not supported for partial faces (face_r_dim < 16)."
```
rtawfik01: why?

PR #1134, opened 2026-01-21
Perf SFPU tests fix - call SFPU for each tile

ndivnicTT:
### Ticket
#1133
#1146 

### Problem description
It was discovered that LLK is measuring better perf results than real results. The reason was that SFPU was invoked only for the 1st tile in the block, which is incorrect behavior.

### What's changed
This PR fixes a critical performance measurement bug where SFPU (Special Function Processing Unit) operations were only being invoked for the first tile in each block, leading to inaccurate performance results. The fix ensures SFPU operations are called for every tile.

Changes:
- Modified SFPU test kernels to call SFPU operations inside the tile loop instead of once per block
- Refactored to use relative tile indices (block_tile) instead of absolute indices (block_start + block_tile)
- Added Rsqrt operation to the Python test suite

### Type of change

- [ ] Documentation update

### Checklist

### DIFF
--- a/tests/sources/eltwise_binary_sfpu_perf.cpp
+++ b/tests/sources/eltwise_binary_sfpu_perf.cpp
@@ -16,10 +16,12 @@
 
 // Globals
 // Globals
-std::uint32_t unp_cfg_context          = 0;
-std::uint32_t pack_sync_tile_dst_ptr   = 0;
-std::uint32_t math_sync_tile_dst_index = 0;
-static constexpr int MAX_TILES_DEST    = is_fp32_dest_acc_en ? 4 : 8;
+std::uint32_t unp_cfg_context                          = 0;
+std::uint32_t pack_sync_tile_dst_ptr                   = 0;
+std::uint32_t math_sync_tile_dst_index                 = 0;
+static constexpr int MAX_TILES_DEST                    = is_fp32_dest_acc_en ? 4 : 8;
+static constexpr ckernel::DstSync DST_SYNC_MODE        = ckernel::DstSync::SyncHalf;
+static constexpr ckernel::BroadcastType BROADCAST_TYPE = ckernel::BroadcastType::NONE;
 
 #ifdef LLK_TRISC_UNPACK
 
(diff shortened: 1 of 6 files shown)


tests/sources/eltwise_binary_sfpu_perf.cpp
```diff
@@ -54,13 +58,13 @@ void run_kernel(const volatile struct RuntimeParams* params)
                     /* iterations*/ params->num_faces * params->TILE_CNT * params->LOOP_FACTOR);
             }
         }
-        else if constexpr (PERF_RUN_TYPE != PerfRunType::PACK_ISOLATE)
+        else if constexpr (PERF_RUN_TYPE != PerfRunType::PACK_ISOLATE) // UNPACK_ISOLATE, L1_TO_L1, L1_CONGESTION
```
skrsmanovicTT: Did we had practice to cover specific run type with return statement instead of just skipping it, so here for pack isolate I guess we would have else branch with return. @sstanisicTT can you please address if there is any difference since in this scenario for PACK_ISOLATE we would execute PROFILER_SYNC() and not return immediately

skrsmanovicTT (changes requested):

tests/sources/eltwise_binary_sfpu_perf.cpp
```diff
@@ -164,49 +170,46 @@ void run_kernel(const volatile struct RuntimeParams* params)
...
                     // Copy from srcA to dest
                     for (int block_tile = 0; block_tile < block_tiles; ++block_tile)
                     {
-                        _llk_math_eltwise_unary_datacopy_<DataCopyType::A2D, DstSync::SyncHalf, is_fp32_dest_acc_en, BroadcastType::NONE, unpack_to_dest>(
-                            block_start + block_tile, formats.math, formats.math);
-                    }
-
-                    // Start SFPU binary operation
-                    _llk_math_eltwise_binary_sfpu_start_<DstSync::SyncHalf>(/* dst_index */ block_start);
+                        _llk_math_eltwise_unary_datacopy_<data_copy_type, DST_SYNC_MODE, is_fp32_dest_acc_en, BROADCAST_TYPE, unpack_to_dest>(
```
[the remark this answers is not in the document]
ndivnicTT: _llk_math_eltwise_unary_datacopy_ - The method has the logic for handling unpack_to_dest. In our tests, we use BroadcastType::NONE exclusively, so when we use unpack_to_dest (and 32b data - handled by .py script) this method unwraps to:
        math_unpack_to_dest_math_ready();
        math::set_dst_write_addr<DstTileShape::Tile32x32, UnpackDestination::DestReg>(dst_index);
        math::math_unpack_to_dest_tile_ready();
 And for BH architecture, there is extra TT_ZEROACC for each face
 
So I would say having it for L1_TO_L1 is "playing by the book" as it can handle unpack_to_dest without actually copying - only for sync purposes.

tests/sources/eltwise_binary_sfpu_perf.cpp
skrsmanovicTT: Any update here ?

tests/sources/pack_untilize_perf.cpp
```diff
@@ -109,15 +109,15 @@ void run_kernel(const volatile struct RuntimeParams* params)
...
-            // FIXME: Currently have no way to mock math for unpack to dest
             for (uint32_t loop = 0; loop < params->LOOP_FACTOR; loop++)
             {
                 for (uint32_t block = 0; block < params->TILE_CNT / BLOCK_CT_DIM; block++)
                 {
                     for (uint32_t block_tile = 0; block_tile < BLOCK_CT_DIM; block_tile++)
                     {
-                        _llk_math_eltwise_unary_datacopy_<DataCopyType::A2D, DstSync::SyncHalf, is_fp32_dest_acc_en, BroadcastType::NONE, unpack_to_dest>(
-                            block_tile, formats.math, formats.math);
+                        math_unpack_to_dest_math_ready();
```
skrsmanovicTT: return back to existin LLK because it is already handled there

PR #1190, opened 2026-01-28
New Fast Approximate Exp Algorithm

sosborne-TT:
### Ticket
[27912](https://github.com/tenstorrent/tt-metal/issues/27912).

### Problem description
The fast-approx exponential function based on the Schraudolph algorithm is used in SDPA. For compute-intensive models like image and video generation it takes a non-negligible amount of time and any improvement in performance will directly speed up those models.

### What's changed
This adds a faster version. On Wormhole: 70.7 cycles / tile  (when called with iterations = 32) vs. 106.7 for the baseline, and essentially the same on Blackhole: 68.6 cycles / tile vs. 104.5 for the baseline. It has some limitations:
 - Inputs <~ -88 will be incorrect (but guaranteed to be negative so if combined with a ReLU, e.g. in the Packer, the result will be correct)
 - Inputs >~ 0.72 will output Exp(0.72), i.e the result will be clamped.

The idea is to use `SETSGN` to set the sign of the output based on the sign of the `STOCHRND` output, and interleave the additional instruction (which is `SHFT2` with `SETSGN` replacing `SHFT` in the `LOADMACRO`) with `LOADMACRO`:
image

This also adds a test for the fast approximate exponential, both the new version and the original one.

The output exactly matches the baseline for values in the range [-88, 0.72]. For values below -88 all outputs are <= 0. Values above 0.72 are clamped to `exp(0.72)`.
exp-bfloat16-pres2

Baseline:
exp-bfloat16-pres2

### Type of change

- [ ] Documentation update

### Checklist

- [x] tt-metal post commit [passes](https://github.com/tenstorrent/tt-llk/actions/runs/21498093062).

### DIFF
--- a/tests/helpers/include/sfpu_operations.h
+++ b/tests/helpers/include/sfpu_operations.h
@@ -73,7 +73,9 @@ void call_sfpu_operation(SfpuType operation, std::uint32_t math_format = 0, floa
             _init_exponential_<APPROX_MODE, FAST_MODE, 0x3F800000 /* exp_base_scale_factor */, CLAMP_NEGATIVE>();
             if constexpr (FAST_MODE && APPROX_MODE && CLAMP_NEGATIVE)
             {
-                static_assert(ITERATIONS == 32); // In this case _calculate_exponential_ always processes 8 iterations.
+                // In this case each call to _calculate_exponential_ processes 8 iterations
+                // and we iterate over 4 faces, so we process 32 iterations overall.
+                static_assert(ITERATIONS == 32);
                 for (int i = 0; i < 4; i++)
                 {
                     _calculate_exponential_<APPROX_MODE, false /* scale_en */, ITERATIONS, FAST_MODE, false /* skip_positive_check */, CLAMP_NEGATIVE>(
@@ -84,7 +86,8 @@ void call_sfpu_operation(SfpuType operation, std::uint32_t math_format = 0, floa
             }
             else if constexpr (FAST_MODE && APPROX_MODE)
             {
-                static_assert(ITERATIONS == 8 || ITERATIONS == 32); // In this case _calculate_exponential_ can process 8 or 32 iterations.
+                // In this case each call to _calculate_exponential_ can process either 8 or 32 iterations.
+                static_assert(ITERATIONS == 8 || ITERATIONS == 32);
                 _calculate_exponential_<APPROX_MODE, false /* scale_en */, ITERATIONS, FAST_MODE, false /* skip_positive_check */, CLAMP_NEGATIVE>(
                     p_sfpu::kCONST_1_FP16B /* exp_base_scale_factor */);
             }


tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_exp.h
```diff
@@ -399,6 +535,117 @@ inline void _init_exponential_()
...
+        //   LREG[13] = B - C = 32500.818359375
+        //   LREG[14] = 15 (shift amount for SFPSHFT2)
+        //
+        // Macro Instructions (backdoor loaded):
+        //   Macro 5: MAD        - compute i = A * x + (B-C)
+        //   Macro 6: STOCHRND   - convert to INT16 (sign-magnitude format)
+        //   Macro 7: SETSGN     - restore sign from STOCHRND result to shifted value
+        //
+        // Macro Sequence Register 0
+        //   Slot 1 (Simple): SETSGN @ delay 5, writes to LREG[16] (staging)
```
fvranicTT: Just an idea. Thinking out loud...
If the caller guarantees inputs are in `[-88, +88]`, the intermediate value is always positive, so we could:

1. Use STOCHRND mode 14 (unsigned) instead of sign-magnitude
2. Eliminate SETSGN entirely
3. Schedule SFPSHFT on the Simple sub-unit within the loadmacro (this is the biggest win in fact)

fvranicTT: Instead of what you have here:
```
Cycle 0: LOAD -> LREG[n%4]           (immediate)
Cycle 1: MAD (delay 0)               A*x + (B-C) -< LREG[n%4]
Cycle 4: STOCHRND (delay 3)          FP32 -> INT16 (sign-magnitude) -> LREG[n%4]
Cycle 5: SHFT (discrete instruction) LREG[n%4] << 15 -> LREG[4]
Cycle 6: SETSGN (delay 5)            Combine sign + shifted value -> LREG[16]
Cycle 7: STORE (delay 6)             LREG[16] -> Dst
```
I think I get the idea with SETSGN. For out-of-range negative inputs (x < -88), the intermediate value is negative. SETSGN preserves this sign in the final result (producing a negative number that's detectably wrong).

But if the input is already clamped and in valid range, we could make a shorter pipeline per element:
```
Cycle 0: LOAD -> LREG[n%4]           (immediate)
Cycle 1: MAD (delay 0)               A*x + (B-C) -> LREG[n%4]
Cycle 4: STOCHRND (delay 3)          FP32 -> UINT16 -> LREG[n%4]
Cycle 5: SHFT (delay 4)              LREG[n%4] << 15 -> LREG[n%4]
Cycle 6: STORE (delay 5)             LREG[n%4] -> Dst
```

So instead of `SHFT, LM, SHFT, LM, ...` alternating 56 times, you could have `LM, LM, LM, LM, ...` 28 times + draining NOPs (have to do the math, not sure ATM how many of them). This way we only execute LM, which should be much faster.

tests/python_tests/test_eltwise_unary_sfpu.py
```diff
@@ -280,3 +282,91 @@ def eltwise_unary_sfpu(
...
+    dest_acc = DestAccumulation.No
+
+    # Generate custom stimuli with range [-5, 0.7]
+    num_elements = input_dimensions[0] * input_dimensions[1]
+    src_A = torch.rand(num_elements, dtype=torch.bfloat16) * 5.7 - 5.0
+    # Set some values to be large and negative:
+    src_A[0] = -10000
+    src_A[1] = -1000
+    src_A[2] = -200
+    src_A[3] = -100
```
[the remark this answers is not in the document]
sosborne-TT: Since the case with input >~ 0.72 will be incorrect for `clamp_negative = false` and since we are primarily interested in using this function for cases where input <= 0, it's less important to test larger values.

nvelickovicTT (approved):

Issue #1243, opened 2026-02-05
Pack reconfig should be more complete


ryanzhuTT:
Investigate if `reconfig_packer_data_format` is missing reprogramming any registers whose values are expected to be different based on values of parameters passed to `llk_pack_untilize_hw_configure`

amahmudTT:
We are not trying to introduce tilize/untilize behavior in reconfiguring. LF8 issue has been handled. SETADCXX is currently being moved around as it is being treated as transient. So closing this issue now, may reopen later if anything is needed.

PR #1244, opened 2026-02-05
Increase coverage of pack reconfig

ryanzhuTT:
### Ticket

[Link to Github Issue](https://github.com/tenstorrent/tt-llk/issues/1243)
### Problem description

Changing arguments to/from the defaults that are passed down through llk_pack_hw_configure can change the values that fill certain registers, but not all of these are addressed properly in a reconfig call when these parameters could be different from before. When the reconfig does not reprogram these registers as expected, this can lead to incorrect behaviour.
### What's changed

Changes have been made to accept more args in reconfig functions where needed and modify register values based on those arguments in accordance with hw_configure.
### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

- [ ] [Assert validation](docs/Introduction_to_asserts.md) Complied with assert doc (if applicable)

### DIFF
--- a/tt_llk_wormhole_b0/llk_lib/llk_pack.h
+++ b/tt_llk_wormhole_b0/llk_lib/llk_pack.h
@@ -108,11 +108,11 @@ inline void _llk_pack_reconfig_data_format_(
     const std::uint32_t pack_src_format,
     const std::uint32_t pack_dst_format,
     const std::uint32_t tile_size,
-    const std::uint32_t face_r_dim = FACE_R_DIM,
-    const std::uint32_t num_faces  = 4,
-    const bool partial_face        = false,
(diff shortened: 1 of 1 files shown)


ryanzhuTT:
Many of these will not be used in config in the future, e.g. untilize/tilize, narrow_tile, setadcxx, but are modified here anyways as a temporary solution since those changes are not coming soon.

That said let me know if any of these are entirely unnecessary.
For example I wasn't sure about relu_config, since it is always passed as default of 0 in `llk_pack_hw_configure` and only potentially used in `llk_pack_untilize_hw_configure_disaggregated` through template parameters and a pack_params union, as well as the fact that there seems to be a `llk_pack_relu_config` function that essentially handles that part already

Issue #1235, opened 2026-02-05
[Perf]: Tiny tile matmul performance


atatuzunerTT:
### Description
Performance benchmarks currently only cover standard 32×32 tile matmul operations. There is no performance visibility for matmul operations using reduced tile dimensions (tiny tiles), which are important for workloads that don't require full tile utilization (e.g. Deepseek).

### Requirements
- Add performance tests for matmul operations where input 0 uses reduced row dimensions (1, 2, 4, 8, 16 rows × 32 columns)
- Input matrix B uses standard 32×32 tiles
- Measure performance across all standard isolation modes (L1-to-L1, unpack, math, pack, L1 congestion)
- Parametrize across relevant data formats, math fidelity modes, and destination sync modes

@rtawfik01 fyi

PR #1246, opened 2026-02-06
[Test infra]: Add perf test for tiny-tile matmul and densify test data for real L1 usage

atatuzunerTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-llk/issues/1235)

### Problem description
Tiny-tile matmul lacked dedicated performance tests, and existing matmul tests did not use input/output data layouts that reflect real L1 usage. This made it harder to validate performance and correctness under realistic conditions.

### What's changed
- **Perf test for tiny-tile matmul:** Added `math_matmul_perf.cpp` and `perf_math_matmul.py` to run and measure tiny-tile matmul performance (pack/math isolate and full run types).
- **Densified test data:** Updated stimuli generation and test data layout so input/output data matches real L1 usage, improving test fidelity and enabling meaningful perf gains.
- **Clearer naming:** Renamed matmul input representation for accuracy (e.g. unpack/math test and sweep helpers).
- **Full test coverage:** Fixed restructuring issues in `test_math_matmul.py`, `test_unpack_matmul.py`, and related helpers so tiny-tile and standard matmul tests have full coverage.

**Impact:** Enables reliable perf benchmarking of tiny-tile matmul and aligns tests with actual L1 data layout.

### Type of change

- [ ] Documentation update
- [x] Refactoring

### DIFF
--- a/tests/python_tests/helpers/matmul_sweep.py
+++ b/tests/python_tests/helpers/matmul_sweep.py
@@ -401,11 +401,7 @@ def sweep_matmul(
         )
 
         for dest_acc in dest_acc_modes:
-            if (  # Remove invalid variants
-                dest_acc == DestAccumulation.No
-                and fmt.input_format == DataFormat.Float16_b
-                and fmt.output_format == DataFormat.Float16
-            ):
+            if is_dest_acc_needed(fmt) and dest_acc == DestAccumulation.No:
                 continue
 
             max_tiles = 4 if dest_acc == DestAccumulation.Yes else base_max_tiles
@@ -476,11 +472,7 @@ def sweep_tiny_tiles_matmul(
         base_max_tiles = 8 if dest_sync == DestSync.Half else 16
         for fmt in formats_list:
             for dest_acc in dest_acc_modes:
-                if (  # Remove invalid variants
-                    dest_acc == DestAccumulation.No
-                    and fmt.input_format == DataFormat.Float16_b
-                    and fmt.output_format == DataFormat.Float16
-                ):
+                if is_dest_acc_needed(fmt) and dest_acc == DestAccumulation.No:
                     continue
 
                 for stochastic_mode in all_stochastic_modes:
--- a/tests/python_tests/perf_math_matmul.py
+++ b/tests/python_tests/perf_math_matmul.py
@@ -21,7 +21,6 @@ from helpers.test_variant_parameters import (
     DEST_INDEX,
     DEST_SYNC,
     IN_TILE_DIMS,
-    INPUT_DIMENSIONS,
     LOOP_FACTOR,
     MATH_FIDELITY,
     NUM_FACES,
@@ -127,7 +126,6 @@ def test_perf_math_matmul(
         formats,
         run_types,
         templates=[
-            INPUT_DIMENSIONS(in0_dimensions, in1_dimensions),
             MATH_FIDELITY(math_fidelity),
             DEST_SYNC(matmul_config.dest_sync),
             THROTTLE_LEVEL(throttle),
--- a/tests/sources/math_matmul_perf.cpp
+++ b/tests/sources/math_matmul_perf.cpp
@@ -30,10 +30,10 @@ void run_kernel(const volatile struct RuntimeParams* params)
     {
         ZONE_SCOPED("INIT")
         _llk_unpack_hw_configure_<is_fp32_dest_acc_en>(
-            formats.unpack_src,
-            formats.unpack_src,
-            formats.unpack_dst,
-            formats.unpack_dst,
+            formats.unpack_A_src,
+            formats.unpack_B_src,
+            formats.unpack_A_dst,
+            formats.unpack_B_dst,
             in1_tile_r_dim < FACE_R_DIM ? in1_tile_r_dim : FACE_R_DIM,
             in0_tile_r_dim < FACE_R_DIM ? in0_tile_r_dim : FACE_R_DIM,
             num_faces_B, // In1
@@ -70,8 +70,8 @@ void run_kernel(const volatile struct RuntimeParams* params)
                 for (std::uint32_t j = 0; j < KT_DIM; j++)
                 {
                     _llk_unpack_AB_matmul_<>(
-                        L1_ADDRESS(buffer_A[0]),
-                        L1_ADDRESS(buffer_B[0]),
+                        L1_ADDRESS(params->buffer_A[0]),
+                        L1_ADDRESS(params->buffer_B[0]),
                         j,
                         j * CT_DIM,
                         TILE_SIZE_UNPACK_A,


tests/python_tests/helpers/matmul_sweep.py
```diff
@@ -115,15 +191,15 @@ def generate_matmul_dimension_combinations(
 
 def generate_matmul_tiny_tiles_combinations(max_tiles: int) -> List[tuple]:
```
skrsmanovicTT: nitpick: Add function description header

tests/python_tests/perf_math_matmul.py
```diff
@@ -0,0 +1,172 @@
...
+    MATMUL_FORMATS,
+    DEST_ACC_MODES,
+    STOCHASTIC_ROUNDING_MODES,
+    DEST_SYNC_MODES,
+    math_matmul=True,
+)
+
+ALL_TEST_PARAMS = list(
+    chain(
+        # Regular matmul combinations with all throttle levels
```
skrsmanovicTT: We should remove everything that is copy/paster from functional test as this is tiny tiles only perf test

tests/python_tests/perf_math_matmul.py
```diff
@@ -0,0 +1,172 @@
...
+    TILE_COUNT,
+    UNPACK_TRANS_FACES,
+    UNPACK_TRANS_WITHIN_FACE,
+)
+
+MATMUL_FORMATS = input_output_formats(
+    [
+        DataFormat.Float16_b,
+        DataFormat.Float16,
+        DataFormat.Float32,
```
skrsmanovicTT: what about bfp8 ?

tests/python_tests/perf_math_matmul.py
```diff
@@ -0,0 +1,172 @@
...
+        "sources/math_matmul_perf.cpp",
+        formats,
+        run_types,
+        templates=[
+            INPUT_DIMENSIONS(in0_dimensions, in1_dimensions),
+            MATH_FIDELITY(math_fidelity),
+            DEST_SYNC(matmul_config.dest_sync),
+            THROTTLE_LEVEL(throttle),
+            TILE_COUNT(variant_tile_count),
+            NUM_FACES(
```
skrsmanovicTT: I would suggest a bit different approach here. I know that our API/lib is pain point with this which in0/in1 is passed to which hw soruce register and arguement switching is happening on several places along the call trace (compute api layer in metal, llk layer in metal, _llk layer in mop configuration etc...)

My suggestion would be to keep tests clean meaning in0 is always in0, in1 is always in1 for all parameters and then handle swap of parameters in cpp test kernel implementation. With this approach test should remain unchanged and with clear and understandable mapping of inputs like for example anything above compute api level in metal should be.
Later if we hopefully fix this mess of how in parameters are passed to llk lib we could update cpp test kernel to match that new layout as well.

skrsmanovicTT (commented):
@atatuzunerTT let's sync offline if there is any question, I have few proposals there and had some general questions

tests/python_tests/perf_math_matmul.py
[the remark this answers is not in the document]
atatuzunerTT: I agree. I'm creating a plan to restructure the matmul sweeps for both regular and tiny tile matmuls. I'll include this point there

tests/python_tests/perf_math_matmul.py
atatuzunerTT: I left it like this since our goal is to later unify regular and tiny tile matmuls under a single file. The unification should follow a similar approach for both functional and perf tests. However, I'm not sure if the perf suite has enough BW to run all matmul combinations at the same time since perf tests seem to run for longer.

tests/sources/math_matmul_perf.cpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+        ZONE_SCOPED("INIT")
+        _llk_unpack_hw_configure_<is_fp32_dest_acc_en>(
+            formats.unpack_src,
+            formats.unpack_src,
+            formats.unpack_dst,
+            formats.unpack_dst,
+            in1_tile_r_dim < FACE_R_DIM ? in1_tile_r_dim : FACE_R_DIM,
+            in0_tile_r_dim < FACE_R_DIM ? in0_tile_r_dim : FACE_R_DIM,
+            num_faces_A, // In1/SrcA
```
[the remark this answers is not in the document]
atatuzunerTT: We also need to swap the `TILE_SIZE_UNPACK_*` parameters. I have a follow-up PR that addresses this: https://github.com/tenstorrent/tt-llk/pull/1265

tests/python_tests/perf_math_matmul.py
```diff
@@ -0,0 +1,172 @@
...
+                matmul_config.tile_dimensions.kt_dim,
+            ),
+            IN_TILE_DIMS(
+                matmul_config.tile_dimensions.in0_tile_r_dim,
+                matmul_config.tile_dimensions.in0_tile_c_dim,
+                matmul_config.tile_dimensions.in1_tile_r_dim,
+                matmul_config.tile_dimensions.in1_tile_c_dim,
+            ),
+            DEST_INDEX(matmul_config.dst_index),
+            LOOP_FACTOR(1024),
```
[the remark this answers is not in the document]
atatuzunerTT: There seems to be non-negligible difference in perf. I attached the results in that issue.

tests/python_tests/perf_math_matmul.py
skrsmanovicTT: Ok great. Please just consider existing normal tile mamtul perf test once you start unifying things. There are some differences compared to functional one so we should close all gaps. Thanks for effort.

tests/python_tests/helpers/matmul_sweep.py
```diff
@@ -74,6 +76,80 @@ class MatmulConfig:
...
+        num_faces_in1: Number of faces for in1 (SrcA)
+        is_in0_horizontal: For 2-face mode, whether in0 uses horizontal layout (f0,f1)
+
+    Returns:
+        Number of output faces (1, 2, or 4)
+    """
+    # Get dimensions for in0
+    a_rows, _ = get_face_dimensions(num_faces_in0, is_horizontal=is_in0_horizontal)
+
+    # For 2-face mode, in1 uses opposite layout of in0
```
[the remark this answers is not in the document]
atatuzunerTT: You're right, `is_in1_horizontal` is set to `True` for any number of faces (e.g. 1 or 4) of in1, but `get_face_dimensions` doesn't use this parameter for those cases. The only case where this parameter is relevant is when there are 2 faces, which is to make sure that matmul dimensions align. So it should be fine

fvranicTT:
> Agreed. I believe bfp4 is currently not supported in our test infra, but I created [this issue](https://github.com/tenstorrent/tt-llk/issues/1329) to track the float16-bfp8 case.

@atatuzunerTT I have an old [branch](https://github.com/tenstorrent/tt-llk/tree/filip/bfp4) to support bfp4, but I didn't finish it (still doesn't work).

rtawfik01 (approved):

sstanisicTT:
Great PR, just wondering, do we have any plots that relate before/after impact of densifying the tiles

atatuzunerTT:
I should've made what densification means clearer, but basically, it is more so for the functional tests: it changes how stimuli is generated so that it matches the expectation of HW with tiny tiles. For perf, this shouldn't make a difference since we don't do accuracy checks

tests/sources/unpack_matmul_test.cpp
```diff
@@ -105,22 +104,39 @@ void run_kernel(const volatile struct RuntimeParams *params)
 {
 #ifdef ARCH_BLACKHOLE
     _llk_pack_hw_configure_<is_fp32_dest_acc_en, false, false>(
-        formats.pack_src, formats.pack_dst, TILE_SIZE_PACK, FACE_R_DIM, TILE_C_DIM, params->num_faces, params->PARTIAL_FACE_PACK);
-    _llk_pack_init_<false, false, false>(formats.pack_dst, FACE_R_DIM, TILE_C_DIM, params->num_faces, false /* partial_face parameter is unused on BH */);
-    _llk_pack_dest_init_<DstSync::SyncHalf, is_fp32_dest_acc_en>();
+        formats.pack_src,
+        formats.pack_dst,
+        TILE_SIZE_PACK,
```
ajankovicTT: These formats won't compile on the latest main, you need to write it like this:
```
formats.unpack_A_src,
formats.unpack_B_src,
formats.unpack_A_dst,
formats.unpack_B_dst,
```

@uvelimirovicTT enahnced our infra formats to allow for setting different formats for srcA and srcB

I saw @skrsmanovicTT commented already on this matter, is it resolved?

PR #1256, opened 2026-02-08
fix: remedy some perf code smells

fvranicTT:
### Problem description
- Issue 1: Replaced `results.groupby("marker").first().reset_index()` with `reduce + pd.merge(..., on="marker", how="outer", validate="1:1")`. This merges the per-run-type stats DataFrames explicitly on the marker column. The `validate="1:1"` argument will raise a `MergeError` if any marker appears more than once in either side, which was previously a silent data loss. The `how="outer"` ensures markers that appear in only some run types are still preserved.
- Issue 2: Added two guards to the `ZONE_END` handling in `_parse_thread`:
1. Empty stack check, which raises `ValueError` with the marker name, id, thread, and timestamp if a `ZONE_END` arrives with no pending `ZONE_START`. Previously this would crash with a bare `IndexError` and was hard to track.
2. Marker ID validation: after popping, verifies that the popped `ZONE_START`'s `marker_id` matches the `ZONE_END`'s `marker_id`. Raises `ValueError` with both marker names and IDs on mismatch, catching corrupted buffer data or nested zone pairing errors.

### What's changed
Added fixes for the above mentioned issues.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

- [ ] [Assert validation](docs/Introduction_to_asserts.md) Complied with assert doc (if applicable)

### DIFF
--- a/tests/python_tests/helpers/fuser_config.py
+++ b/tests/python_tests/helpers/fuser_config.py
@@ -155,7 +155,7 @@ class FuserConfig:
                 test_config.generate_variant_hash()
                 test_config.build_elfs()
 
-                for run_idx in range(run_count):
+                for run_index in range(run_count):
                     elfs = test_config.run_elf_files(location)
                     wait_for_tensix_operations_finished(elfs, location)
 
@@ -172,7 +172,7 @@ class FuserConfig:
                     ]
                     profiler_data = Profiler._parse_buffers(buffer_data, meta)
                     # Tag profiler data with run index for proper L1-to-L1 pairing
-                    profiler_data.df["run_index"] = run_idx
+                    profiler_data.df["run_index"] = run_index
                     runs.append(profiler_data)
 
                 get_stats = Profiler.STATS_FUNCTION[run_type]
--- a/tests/python_tests/helpers/perf.py
+++ b/tests/python_tests/helpers/perf.py
@@ -93,7 +93,7 @@ class PerfReport:
             mask & (frame[column] == value)
             for frame, mask in zip(self._frames, self._masks)
         ]
-        return PerfReport(frames=self._frames.copy(), masks=mask_chain)
+        return PerfReport(frames=self._frames, masks=mask_chain)
 
     def marker(self, marker: str) -> "PerfReport":
         """Filter: Marker"""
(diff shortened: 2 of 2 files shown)


tests/python_tests/helpers/perf.py
```diff
@@ -373,8 +374,12 @@ def run(self, perf_report: PerfReport, run_count=2, location="0,0"):
             get_stats = Profiler.STATS_FUNCTION[run_type]
             results.append(get_stats(ProfilerData.concat(variant_raw_data)))
 
-        results = pd.concat(results, ignore_index=True)
```
fvranicTT: Issue:
- `pd.concat()` stacks dataframes vertically, potentially creating duplicate rows with the same marker
- `groupby("marker").first()` silently takes the first occurrence when duplicates exist, hiding data corruption
- if two different run types produced stats for the same marker name, one would be silently discarded

Proposed fix:
- `pd.merge()` with `validate="1:1"` ensures each marker appears exactly once in each result, see [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html#pandas-dataframe-merge) for more
- if duplicates exist, the merge will fail loudly with a validation error instead of silently dropping data

tests/python_tests/helpers/fuser_config.py
```diff
@@ -154,7 +155,7 @@ def run(self, worker_id="master", location="0,0", run_count=2):
                 test_config.generate_variant_hash()
                 test_config.build_elfs()
 
-                for _ in range(run_count):
+                for run_idx in range(run_count):
```
sstanisicTT: nit: `run_index`

tests/python_tests/helpers/perf.py
```diff
@@ -358,7 +359,7 @@ def run(self, perf_report: PerfReport, run_count=2, location="0,0"):
             self.runtimes = runtimes
             self.generate_variant_hash()
             variant_raw_data = []
-            for _ in range(run_count):
+            for run_idx in range(run_count):
```
sstanisicTT: nit: `run_index`

fvranicTT: Fixed.

Issue #873, opened 2025-11-24
Tiny Tile Matmul Double Accumulation (Potential) Bug


skotaracTT:
Tile shape: [X, 32] with X ∈ {1, 2, 4, 8}
Sync mode: DestSync.Full
Condition: dst_index > 3
For each case that fails, the result values are exactly double what we expect in golden. My guess is that hardcoded DstTileShape::Tile32x32 in the math_matmul function is incompatible for tiny tiles and introduces incorrect memory address increments (<< 6 bits) leading to:

Address overflow when dst_index > 3
Memory overlap, causing double accumulation
Invalid memory access in some cases

Issue #697, opened 2025-09-25
Incorrect Matmul with Tiny Input Tile and Multi-Tile Output


skotaracTT:
#### Matmul Fails with Tiny Tiles and Multiple Regular Tiles
Matrix multiplication fails when using tiny tiles in combination with larger output tiles. Specifically:
- Matmul with shapes like 8x32 and 32x32 works correctly (first two faces are accurate).
- Matmul with shapes like 8x32 and 32x64 produces incorrect results:
     - First tile produces correct output.
     - Subsequent tiles (next two faces f0, f1) are invalid, there is no clear or consistent pattern in the erroneous output.

Reproducible Combinations
This issue is consistently reproducible for all combinations where:
Left-hand matrix: [X, 32] with X ∈ {1, 2, 4, 8}
Right-hand matrix: [32, Y] with Y ∈ {64, 96, 128, 160, 192, 224, 256}

pmilenkovicTT:
The PR to fix the [#29890](https://github.com/tenstorrent/tt-metal/issues/29890) has been merged. 
@skotaracTT could you try running the tests again to figure out if the issue still occurs? 
Bear in mind our talk with @rtawfik01, where we discussed the supported tile dimensions for tiny tiles.

skrsmanovicTT:
https://github.com/tenstorrent/tt-llk/pull/1265 Should be fixed with this PR

PR #1265, opened 2026-02-10
[Test infra]: Enabling multi-tile matmul with tiny tiles

atatuzunerTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-llk/issues/697) https://github.com/tenstorrent/tt-llk/issues/873

### Problem description
We want to enable multi-tile output matmuls for tiny tiles. Two bugs were blocking this:
1. **Stimuli generation:** When multiple output tiles were used, face matmul stimuli were generated incorrectly (layout/format did not match real L1 usage).
2. **Golden comparison:** The golden check used wrong indices when comparing per-tile results, so multi-tile outputs were validated incorrectly.

### What's changed
- **stimuli_generator.py:** Reworked `generate_face_matmul_data` so multi-tile output is correct. Data is built in tilized form (rt, ct, 4, 16, 16) using identity/incrementing face tensors, then converted to row-major (rows, cols), matching real L1 layout.
- **test_math_matmul.py / test_unpack_matmul.py:** Fixed per-tile golden comparison by using the correct stride (1024 elements per 32×32 tile) when indexing into the golden tensor, so each output tile is compared against the right slice.
- **matmul_sweep.py:** Removed workarounds that forced single-tile output for tiny tiles (output_tile_cnt=1, dst_index cap). Temporarily disabled dst_index edge-case combinations due to double-acc bug (issue #873), with comments for follow-up.

**Impact:** Multi-tile output matmul for tiny tiles is now supported: stimuli and golden checking are correct, and the sweep no longer restricts to one output tile.

**Note:** This PR is based off of https://github.com/tenstorrent/tt-llk/pull/1246 changes.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### DIFF
--- a/tests/python_tests/helpers/stimuli_generator.py
+++ b/tests/python_tests/helpers/stimuli_generator.py
@@ -197,15 +197,15 @@ def generate_face_matmul_data(
     rt, ct = rows // DEFAULT_TILE_R_DIM, cols // DEFAULT_TILE_C_DIM
     dtype = format_dict[stimuli_format]
 
-    out = torch.empty((rows, cols), dtype=dtype)
-    for i in range(rt):
-        for j in range(ct):
-            tile = torch.rand(DEFAULT_TILE_R_DIM, DEFAULT_TILE_C_DIM, dtype=dtype)
-            masked = _mask_tile(tile, num_faces, not is_matrix_A, face_r_dim)
-            out[
-                i * DEFAULT_TILE_R_DIM : (i + 1) * DEFAULT_TILE_R_DIM,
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_unpack_matmul.py
```diff
@@ -231,7 +231,7 @@ def test_unpack_matmul(math_fidelity, matmul_config, workers_tensix_coordinates)
 
         # Compare each tile separately
         for i in range(tile_cnt):
-            start = i * num_elements_per_tile
+            start = i * 1024
```
fvranicTT: Ditto

atatuzunerTT:
Nightly passes with same failure as main

Issue #1168, opened 2026-01-26
Asserts for data format validation


ndivnicTT:
Problem Statement
Incorrect or incompatible data format usage can lead to invalid pack operations, corrupted results, or runtime failures. For example, attempting to pack a Float32 destination into an Int8 format is not supported and must be explicitly prevented. Assertions are required to enforce format validity and compatibility at both compile‑time and runtime.

PR #1276, opened 2026-02-11
Add LLK_ASSERTs which verify unpack and pack src and dst formats are compatible

ndivnicTT:
### Ticket

#1168
#1167
#1332

### Problem description
Currently, there is no validation for unpacker and packer HW configure, init or reconfigure related to compatibility between src_format and dst_format. In other words, for unpacker between unpack_src_format and unpack_dst_format (for any of 2 unpackers) and same for pack_src_format and pack_dst_format for packer. This can lead up to invalid combinations being used which could potentially lead up to undefined behavior/wrong results.

### What's changed
This PR introduces functions that verify allowed combinations for src and dst formats:
- is_unpacker_to_register_conversion_supported
- is_packer_to_L1_conversion_supported
For unpacker it is important to have in mind these 2 flags:
- is_fp32_dest_acc_en
- unpack_to_dest

While we don't have unpack_to_dest during configure_unpack_AB, and have to workaround by using is_32bit_input(...), when unpack_to_dest is really needed, it will be available during _llk_unpack_A_init_, due to the fact that unpack to dest is performed by unpack A only.

In order to create functions that will validate the compatibility, ISA documentation was used:
1. https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/Unpackers/FormatConversion.md
2. https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/Packers/FormatConversion.md

Notice that both docs are under WormholeB0 directory. For Blackhole, there are no such docs, so the idea was to just copy the same logic that is described in Wormhole docs.

Also, there are couple of perf tests that required skipping in order to fulfill the new checks.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

- [x] [Assert validation](docs/Introduction_to_asserts.md) Complied with assert doc (if applicable)

### DIFF
--- a/tt_llk_blackhole/common/inc/cpack_common.h
+++ b/tt_llk_blackhole/common/inc/cpack_common.h
@@ -200,10 +200,8 @@ __attribute__((noinline)) bool is_packer_combined_late_column_output(const DataF
 /**
  * \brief Returns true if conversion is supported by EARLY packer conversion stage.
  *
- * This checks the Wormhole early conversion matrix only (Dst register format -> intermediate
- * format). For this API, `out_l1` is interpreted as the requested intermediate format code.
- *
- * Ref: Wormhole Packers/FormatConversion.md, "Early format conversion".
+ * Validates supported dst-register to intermediate-format pairs for Blackhole's early conversion
+ * stage. For this API, `out_l1` is interpreted as the requested intermediate format code.
  */
 __attribute__((noinline)) bool is_packer_to_L1_early_conversion_supported(const DataFormat in_reg, const DataFormat out_l1)
 {
@@ -249,9 +247,8 @@ __attribute__((noinline)) bool is_packer_to_L1_early_conversion_supported(const 
 /**
  * \brief Returns true if conversion is supported by LATE packer conversion stage.
  *
- * This checks the Wormhole late conversion matrix only (intermediate/LateFromFormat -> L1).
- *
- * Ref: Wormhole Packers/FormatConversion.md, "Late format conversion".
+ * Validates supported intermediate (LateFromFormat) to L1 pairs for Blackhole's late conversion
+ * stage.
  */
 __attribute__((noinline)) bool is_packer_to_L1_late_conversion_supported(const DataFormat in_reg, const DataFormat out_l1)
 {
@@ -353,9 +350,6 @@ inline void reconfigure_packer_l1_acc(const std::uint32_t pack_l1_acc)
  * This is specifically done in FP32 (Dest) -> BFPxA (L1) case.
  *
  * This implies that exponent thresholding should be reconfigured whenever packer formats change.
- *
- * @see https://github.com/tenstorrent/tt-isa-documentation/tree/main/WormholeB0/TensixTile/TensixCoprocessor/Packers/ExponentThresholding.md
- * (Blackhole thresholding works the same as WormholeB0)
  *
  * @tparam is_fp32_dest_acc_en True when Dest register is FP32.
  * @param pack_dst_format Pack output data format.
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/common/inc/cpack_common.h
```diff
@@ -163,6 +163,291 @@ inline void packer_addr_counter_init()
...
+        //      FP32  (identity)           → Late from FP32 col:      Float32, Float16_b, Float16,
+        //                                                             Lf8, Bfp8_b, Bfp4_b, Bfp2_b,
+        //                                                             Bfp8, Bfp4, Bfp2.
+        //      TF32  (rounding)           → Late from combined col:  adds Tf32 in L1.
+        //      BF16  (rounding/trunc)     → Late from combined col:  same as TF32 path.
+        //      INT32 (bitcast)            → Late from INT32:         Int32.
+        //      INT8  (sign+mag extract)   → Late from INT8/UINT8:    Int8.
+        //      UINT8 (low 8b extract)     → Late from INT8/UINT8:    UInt8.
+        //
+        //    Union of all reachable L1 outputs:
```
skotaracTT: The implementation is ISA-faithful, but the function has grown large and contains redundancy (e.g., Float32 and Int32 enumerate identical L1 outputs). The full union of formats are manually encoded and reachable through early + late conversion per dst format, effectively hardcoding the transitive closure of the two-stage pipeline.
- While correct, this makes the logic harder to maintain. 
- It may be cleaner to model early (dst → intermediate) and late (intermediate → L1) conversions in two separate functions and then compose them, which would make for easier maintenance and better reflect the hardware structure.

tt_llk_wormhole_b0/common/inc/cunpack_common.h
```diff
@@ -204,6 +204,547 @@ inline void enable_int8_fpu_math()
...
+ * Used to determine unpack-to-dest mode and related configuration when both
+ * input and output are 32-bit. Masks low nibble of format codes for comparison.
+ *
+ * \param unpack_src_format Unpacker input (L1) data format.
+ * \param unpack_dst_format Unpacker output (register) data format.
+ * \return true if both formats are Int32 or Float32; false otherwise.
+ */
+inline constexpr bool is_32bit_input(const std::uint32_t unpack_src_format, const std::uint32_t unpack_dst_format)
+{
+    const std::uint32_t input_df  = masked_data_format(unpack_src_format);
```
fvranicTT:
```suggestion
    const auto input_df  = static_cast<DataFormat(masked_data_format(unpack_src_format));
```
And then comparison in line 221 becomes way less verbose:
```cpp
return (input_df == DataFormat::Int32 || input_df == DataFormat::Float32) &&
           (output_df == DataFormat::Int32 || output_df == DataFormat::Float32);
```

tt_llk_blackhole/llk_lib/llk_unpack_tilize.h
```diff
@@ -70,6 +70,10 @@ inline void _llk_unpack_tilize_init_(
     const bool unpack_to_dest =
         (unpack_source_format == DataFormat::UInt32) || (unpack_source_format == DataFormat::Int32) || (unpack_dest_format == DataFormat::Float32);
 
+    LLK_ASSERT(
+        is_unpacker_format_conversion_supported_dest(static_cast<DataFormat>(unpack_src_format), static_cast<DataFormat>(unpack_dst_format), unpack_to_dest),
+        "Unsupported unpacker format conversion.");
```
nvelickovicTT: Do we have capability to also print the actual unpack_src_format and unpack_dst_format values as well? It would be very informative.

Issue #1164, opened 2026-01-24
tensix_sync() is not totally correct.


amahmudTT:
According to the description here about tensix_sync()

https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/BabyRISCV/TTSync.md

The sync should use the value read from the pc_buf_base[1], otherwise instructions can still execute in parallel to the execution of tensix_sync(), probably this is not a major issue but we should still fix it. 

Another issue that we do not have any example of is that, during the execution of tennis_sync(), if a memory load is done from a very specific address we could get a hang. 

We should follow the code example given in the ISA doc.

amahmudTT:
This was fixed with this PR, there is still a hardware issue, which cannot be fixed though (In a certain cycle , when the hardware thinks its not busy, tensix sync may just pass, but this has probably not caused any issue, more details in Marco's video lecture on waveforms). 

Closing this issue.

PR #1285, opened 2026-02-12
fix :bug:: add load_blocking, store_blocking

sstanisicTT:
### Ticket

#1164

### Problem description

Tries to fix the bug described in the issue for wormhole and blackhole, while adding helpers for load and store barriers

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tt_llk_blackhole/common/inc/ckernel.h
+++ b/tt_llk_blackhole/common/inc/ckernel.h
@@ -100,17 +100,13 @@ inline std::uint32_t load_blocking(volatile std::uint32_t *ptr)
 
     // https://github.com/tenstorrent/tt-isa-documentation/tree/main/BlackholeA0/TensixTile/BabyRISCV/MemoryOrdering.md
 
-    // important note: FENCE on Wormhole is a NOP
-    //
-    // this code provides a full barrier for a load transaction by doing the following:
-    // - memory clobber
-    //     - prevent reordering of transactions that occur before the barrier after the barrier by the COMPILER
+    // this code provides a blocking load by doing the following:
     // - issue a LOAD transaction to the address
     //     - actual load that was requested
     // - issue an instruction that requires the data from the LOAD transaction
     //     - block the pipeline until the LOAD transaction completes
     // - memory clobber
-    //     - prevent reordering of transactions that occur after the barrier before the barrier by the COMPILER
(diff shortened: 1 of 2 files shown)


tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -88,26 +89,79 @@ namespace internal
...
+ */
+inline std::uint32_t load_blocking(volatile std::uint32_t *ptr)
 {
-    volatile std::uint32_t foo     = 0;
-    volatile std::uint32_t *fooptr = &foo;
-    // Write to pc buffer to push all writes ahead of us.. otherwise, the pc buffer read can bypass older writes
-    pc_buf_base[1] = foo;
+    std::uint32_t val;
+
+    // https://github.com/tenstorrent/tt-isa-documentation/tree/main/WormholeB0/TensixTile/BabyRISCV/MemoryOrdering.md
```
amahmudTT: Its not a big problem, but you are pointing to wormwhole doc in blackhole code. You could point to the blackhole code

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -88,26 +89,79 @@ namespace internal
...
+        : [ptr] "r"(ptr));
 
-    // Now read -- this read will block until mops are done
-    *fooptr = pc_buf_base[2];
+    asm volatile("" : ::"memory");
+}
+
+inline void tensix_sync()
+{
+    store_blocking(&pc_buf_base[1], 0);
```
amahmudTT: the naming is a bit counterintuitive, the part that does the syncing is actually the load inside the store_blocking function. The store earlier is just to make sure previous stores are done. A better name for the function would be store_and_load_blocking, or may be just have the main code inside here as it was before.

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -88,26 +89,79 @@ namespace internal
 {
 }
 
-inline void tensix_sync()
+/**
+ * @brief Issues a load transaction that will block the core until the transaction is completed
+ * @param ptr address to read from
+ * @return value read from the address
+ */
+inline std::uint32_t load_blocking(volatile std::uint32_t *ptr)
```
amahmudTT: Is this needed at all ? I do not see it being used.

PR #1289, opened 2026-02-12
Solution for WTF error

njokovicTT:
### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tests/helpers/src/trisc.cpp
+++ b/tests/helpers/src/trisc.cpp
@@ -33,12 +33,49 @@ std::uint32_t open_zone_cnt = 0;
 extern const volatile struct RuntimeParams __runtime_args_start[];
 extern void run_kernel(const volatile struct RuntimeParams* params);
 
+// Debug breadcrumb addresses in L1 for diagnosing soft resets
+// Each TRISC gets its own address to track execution state
+#if defined(LLK_TRISC_UNPACK)
+constexpr std::uint32_t DEBUG_STATE_ADDR    = 0x64FF8;
+constexpr std::uint32_t DEBUG_LOOP_CNT_ADDR = 0x65004; // Loop counter for trisc0
+#elif defined(LLK_TRISC_MATH)
+constexpr std::uint32_t DEBUG_STATE_ADDR    = 0x64FFC;
+constexpr std::uint32_t DEBUG_LOOP_CNT_ADDR = 0x65008; // Loop counter for trisc1
+#elif defined(LLK_TRISC_PACK)
+constexpr std::uint32_t DEBUG_STATE_ADDR    = 0x65000;
+constexpr std::uint32_t DEBUG_LOOP_CNT_ADDR = 0x6500C; // Loop counter for trisc2
+#endif
+
+// Debug marker values to track execution progress
+// High nibble encodes restart count (0xN), low 3 nibbles encode state
+constexpr std::uint32_t TRISC_MARKER_ENTRY         = 0xDEAD0001;
+constexpr std::uint32_t TRISC_MARKER_AFTER_SETUP   = 0xDEAD0002;
+constexpr std::uint32_t TRISC_MARKER_AFTER_UNRESET = 0xDEAD0003;
+constexpr std::uint32_t TRISC_MARKER_BEFORE_INIT   = 0xDEAD0004;
+constexpr std::uint32_t TRISC_MARKER_AFTER_INIT    = 0xDEAD0005;
+constexpr std::uint32_t TRISC_MARKER_BEFORE_KERNEL = 0xDEAD0006;
+constexpr std::uint32_t TRISC_MARKER_AFTER_KERNEL  = 0xDEAD0007;
+constexpr std::uint32_t TRISC_MARKER_AFTER_SYNC    = 0xDEAD0008;
+constexpr std::uint32_t TRISC_MARKER_DONE          = 0xDEAD0009;
+
 int main()
 {
-#if defined(LLK_TRISC_UNPACK) && defined(LLK_BOOT_MODE_TRISC)
+    volatile std::uint32_t* const debug_state = reinterpret_cast<volatile std::uint32_t*>(DEBUG_STATE_ADDR);
+    volatile std::uint32_t* const loop_cnt    = reinterpret_cast<volatile std::uint32_t*>(DEBUG_LOOP_CNT_ADDR);
+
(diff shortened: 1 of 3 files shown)


tests/helpers/ld/memory.blackhole.ld
```diff
@@ -48,6 +49,7 @@ MEMORY
     TRISC2_CODE             : ORIGIN = ORIGIN(TRISC1_LOADER_INIT_MEM) + LENGTH(TRISC1_LOADER_INIT_MEM), LENGTH = 16K
     TRISC2_LOADER_INIT_MEM  : ORIGIN = ORIGIN(TRISC2_CODE) + LENGTH(TRISC2_CODE), LENGTH = TRISC_LOCAL_MEM_LENGTH
 
+    MAILBOXES               : ORIGIN = MAILBOXES_START, LENGTH = 512
```
fvranicTT: We need this for WH too

tests/helpers/src/trisc.cpp
```diff
@@ -61,11 +63,15 @@ int main()
...
+    ckernel::store_blocking(mailbox, 0x02);
+
     {
         ZONE_SCOPED("KERNEL")
         run_kernel(__runtime_args_start);
         ckernel::tensix_sync();
     }
 
-    *mailbox = ckernel::KERNEL_COMPLETE; // 0x1
+    ckernel::store_blocking(mailbox, 0x03);
```
fvranicTT: Can we use named constants, like KERNEL_COMPLETE? Also, we have 2 consecutive writes here, 0xFF should be enough.

tests/python_tests/helpers/device.py
```diff
@@ -225,37 +216,70 @@ def handle_if_assert_hit(elfs: list[str], core_loc="0,0", device_id=0):
         raise LLKAssertException(temp_stack_traces)
 
 
+def make_sure_all_out_of_reset(location: str = "0,0", backoff=0.01):
```
fvranicTT:
```python
def wait_for_cores_out_of_reset(location: str = "0,0", timeout=0.01):
    full_mask = get_soft_reset_mask(ALL_CORES)
    register_store = get_register_store(location, 0)
    end_time = time.time() + timeout
    
    while time.time() <= end_time:
        soft_reset = register_store.read_register("RISCV_DEBUG_REG_SOFT_RESET_0")
        
        if soft_reset & full_mask == 0:
            return
        
        time.sleep(0.0001)  # avoid busy-waiting with 100us sleep
    
    # Timeout occurred
    soft_reset = register_store.read_register("RISCV_DEBUG_REG_SOFT_RESET_0")
    raise Exception(
        f"Triscs not started within {timeout}s! BRISC ebreak: {is_assert_hit('BRISC', location)}. Reset reg: {hex(soft_reset)}"
    )
```
1. wait_for_cores_out_of_reset - clearly indicates it's a blocking op
2. cache register_store
3. this loop structure is a bit more efficient, we check time first
4. backoff is in fact timeout
5. avoid busy waiting
6. return provides a cleaner control flow
7. read register one final time on timeout to make sure it's the final state.

tt_llk_blackhole/common/inc/ckernel.h
```diff
@@ -88,26 +85,80 @@ namespace internal
...
+    // - issue a STORE transaction to the address
+    //     - actual store that was requested
+    // - issue a LOAD transaction to the address
+    //     - must complete after the STORE transaction
+    // - issue an instruction that requires the data from the LOAD transaction
+    //     - block the pipeline until the LOAD transaction completes, ensuring that the STORE is complete
+    // - memory clobber
+    //     - prevent reordering of transactions that occur after the barrier before the barrier by the COMPILER
+
+    asm volatile(
```
fvranicTT: Same problem, probably the same fix:
```cpp
asm volatile(
    "fence\n\t"
    "sw %[val], (%[ptr])\n\t"
    "lw %[val], (%[ptr])\n\t"
    "and x0, x0, %[val]"
    : [val] "+r"(val)
    : [ptr] "r"(ptr)
    : "memory");
```

tests/helpers/src/brisc.cpp
```diff
@@ -4,16 +4,49 @@
...
 #include <cstdint>
 
+#include "ckernel.h"
+#include "llk_assert.h"
+
 #ifdef LLK_BOOT_MODE_BRISC
 #include "boot.h"
 #endif
 
+enum class CommandState : std::uint32_t
```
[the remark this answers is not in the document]
njokovicTT: We will not use this, I'm removing this

tt_llk_blackhole/common/inc/ckernel.h
sstanisicTT: I am think it's okay to change this, just a slight note behind the intention so that it stays documented.

Note: I removed the memory clobber before the fence and the blocking store. Even if we prevent the compiler from reordering earlier **stores** past the blocking store, we still can’t prevent the hardware from reordering the resulting memory transactions.

To illustrate:

1. A store instruction (I1) enters the Load/Store Queue (LSQ).
2. A write transaction (T1) is issued to L1.
3. I1 exits the LSQ as soon as T1 is issued (stores exit the LSQ when the transaction is sent, not when the write completes).
4. The fence (I2) and the blocking store (I3) enter the LSQ.
5. A second write transaction (T2) is issued to L1.
6. If T1 and T2 target different addresses, they are independent. The memory system may execute T2 before T1.

Because of this, the sequence behaves like a *half barrier*. We can guarantee that stores appearing **after** the blocking store won’t execute before it, but we cannot guarantee that stores appearing **before** it won’t execute after it at the hardware level.

That’s why the function was renamed to `store_blocking` instead of `store_barrier` - it does not provide full barrier semantics.

For the same reason, the explicit memory clobber is placed *after* the sequence rather than inside it.

Now that I think about it, this behavior is fairly unintuitive and probably deserves more detailed documentation.

tests/helpers/ld/memory.blackhole.ld
```diff
@@ -10,6 +10,7 @@ TRISC_LOCAL_MEM_LENGTH = 4K;
 NCRISC_LOCAL_MEM_LENGTH = 8K;
 
 L1_BASE = 0x0;
+MAILBOXES_START = 0x1FFC0;
```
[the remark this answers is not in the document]
ajankovicTT: With latest changes, addresses are checked and corrected

PR #1320, opened 2026-02-20
[test infra] Nightly tests fix

ajankovicTT:
### Problem description

Nightly tests were failing for some times. There were few issues.
- Variants generated for matmul tests weren't properly sanitized and some were hitting LLK asserts.
- Log1p mathop for test_eltwise_unary_sfpu hits the same compiler bug as many more operations in that tests.
- Coverage processing was taking dubious amounts of time per worker
- Golden stimuli was being generated during the compilation phase which induced a lot of unnessesary work in nightly tests.
- Nightly is still flaky, further testing is required to make it more robust. 

### What's changed

- sanitization was added to matmul sweep functions
- Log1p mathop was added to the lists of operations that should be skipped when testing with coverage
- gcda streams generated by test binaries on device arr merged for all variants sharing elfs before they are converted to .info files, yielding ~100x time reduction.
- Made a dummy_golden_generator function that gets called instead of real golden generator when `TestConfig.Mode == TestMode.Compilation`. This yielded 10x time reduction on biggest nightly tests.
- Tuned `use_durations` off for nightly and PR runs, to more evenly spread work amongs workers. 

### Type of change

- [ ] Documentation update
- [x] Refactoring

### Checklist

### DIFF
--- a/tests/python_tests/test_sfpu_binary.py
+++ b/tests/python_tests/test_sfpu_binary.py
@@ -112,7 +112,13 @@ def test_sfpu_binary_add_top_row(formats, dest_acc, mathop, workers_tensix_coord
         1,
         input_dimensions,
         formats.output_format,
-    ).view(input_dimensions)
+    )
+
+    golden_tensor = (
+        golden_tensor.view([32, 32])
+        if golden_tensor.shape == torch.Size([1024])
+        else golden_tensor.view(input_dimensions)
+    )
 
     configuration = TestConfig(
         "sources/sfpu_binary_test.cpp",


tests/python_tests/test_math_matmul.py
```diff
@@ -125,7 +124,7 @@ def test_math_matmul(
     )
 
     src_B_golden = src_B
-    if transpose == Transpose.Yes:
+    if transpose == Transpose.Yes and TestConfig.MODE != TestMode.PRODUCE:
```
fvranicTT: Same here.

PR #1356, opened 2026-02-25
Add topk stable sort as a parameter for testing.

uvelimirovicTT:
### Problem description
topk test had hardcoded STABLE_SORT=false

### What's changed
This PR enables us to set stable_sort to true if needed.

Currently stable sort has issues but once that's fixed, users should be able to test it with this updated topk test.


fvranicTT:
> Currently stable sort has issues but once that's fixed, users should be able to test it with this updated topk test.

@uvelimirovicTT are you referring to the test or to the kernel?

Issue #1350, opened 2026-02-24
FP8_E4M3 datacopy test fails when extended to this data format


fvranicTT:
I tried adding support for FP8_E4M3 data format in testing harness using ml_dtypes package.
Current work is captured on a branch [`filip/fp8-e4m3`](https://github.com/tenstorrent/tt-llk/compare/main...filip/fp8-e4m3)

What I'm observing in datacopy test is that the correct data is visible in L1 input buffer, but output buffer contains different data.

Example:
```bash
2026-02-24 17:35:28 | DEBUG    | helpers.stimuli_config:write:343 - 
────────────────────────────────────────────────────
  L1 layout @ 0,0
  A    0x00021000  4 × 1024 B
  B    0x00022000  4 × 1024 B
  Res  0x00023000
────────────────────────────────────────────────────
2026-02-24 17:35:28 | DEBUG    | helpers.stimuli_config:collect_results:476 - Reading Res  0x00023000 ← 4096 B
```

```cpp
// INPUT DATA
[4B MODE] noc:0 device:0 loc:2-2 (0,0) > brxy 0,0 0x00021000 16 --format hex32
2-2 (0,0) : 0x00021000 (64 total bytes)
(l1) : 0x00021000 (64 bytes)
0x00021000:  33343234  3134212e  2233302b  211f2536
0x00021010:  2a343236  39382c33  23382c36  38323838
0x00021020:  2838372e  31383435  27302739  2d322f38
0x00021030:  2d32382c  3633342b  37383521  35323422

// OUTPUT DATA
[4B MODE] noc:0 device:0 loc:2-2 (0,0) > brxy 0,0 0x00023000 16 --format hex32
2-2 (0,0) : 0x00023000 (64 total bytes)
(l1) : 0x00023000 (64 bytes)
0x00023000:  393a393a  383a3037  31393835  302f323b
0x00023010:  353a393b  3c3c3639  313c363b  3c393c3c
0x00023020:  343c3b37  383c3a3a  3338333c  3639373c
0x00023030:  36393c36  3b393a35  3b3c3a30  3a393a31
```
Raw hex data seems to be correctly interpreted in Python, I checked raw bytes manually using https://fvranictt.github.io/ieee754-converter/

pgardnerTT:
Fixed the dest_acc=YES cases, but there's still issues with tilize. It's potnentially a HW limitation, will investigate further later. I don't think right now we need it.
[...]

PR #1434, opened 2026-03-09
Add support for Fp8_e4m3

pgardnerTT:
### Ticket
https://github.com/tenstorrent/tt-llk/issues/1350
https://github.com/tenstorrent/tt-metal/issues/18157

### Problem description
Add support for Fp8_e4m3

### What's changed
Add configuration for fp8 e4 register for upk and pck inits
Add WA for float16b->fp8 packer conversion (pack_src_fmt will be float16_b but HW needs to be set to float16). pck_src_fmt is always float16 for Fp8 because the HW understands it as the fmt which comes after the gasket. So float16_b DEST -> Gasket with 10b set -> float16 (pack_src) -> PCK -> fp8 (pack_dst)

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist


rtawfik01:
I agree with @fvranicTT , I prefer atleast one test with this PR.

Also good news, Rakesh has confirmed that tilize and untilize should work with FP8 for both BH and Quasar @nvelickovicTT @pgardnerTT , so those tests can be added too.

PR #1462, opened 2026-03-13
Update LLK_ASSERT macro definition

ndivnicTT:
### Problem description
With current implementation of LLK_ASSERT, when used in metal by enabling only TT_METAL_LLK_ASSERTS=1, after assert is hit and tt-triage ran to print callstacks, it can happen that printed callstack is not the actuall callstack of the assert. In other words, the wrong callstack gets printed.

### What's changed
The change is to delete UNREACHABLE in LLK_ASSERT definition.
There were several options taken into consideration:
- ebreak with FILE and LINE
- trap instruction
- ebreak + no UNREACHABLE

The first two options reduce assembly size a bit, however, the first option increases code size significantly, and second option can help extract right callstack but arguments are missing. The last option is working fine and we can extract callstack and arguments with a bit of extra assembly size.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist


tt-vjovanovic:
Somewhat related to PR, it reminded me on an old topic, when will you have "common" directory that will be shared between architectures?

PR #1463, opened 2026-03-13
[infra. ] Speed of light flag for perf tests and -Werror flag for compilation

ajankovicTT:
### Problem description
Performance tests produced somewhat skewed results, due to compiler not being able to inline all of the code. On tile loops greater than 3, runtime parameter enabled results were in 0.01 cycle of margin from fully compiled ones, but `INIT` and `KERNEL` perf markers were almost double their fully compiled counterparts. This wan't accepatble due to us wanting the _speed of light_ results &rarr; what is the lowes execution time our hardware is capable of.

### What's changed
To achieve this, we introduced a new `pytest` flag called `--speed-of-light`. It is only relevant for perf tests for now. It converts all runtime primitives (formats, stimuli addresses, runtime arguments) into compile time ones. All their values are placed in varaint's `build.h` in a `constexpr` form. This greatly increases compile-time of every tests due to the fact that there aren't any reductions. Furthermore, when this flag is ommited from the `pytest`, runtime primitives stay runtime. This is done by having:
```
#ifndef SPEED_OF_LIGHT
    const std::uint32_t LOOP_FACTOR = params.LOOP_FACTOR;
    std::uint32_t CT_DIM            = params.CT_DIM;
    std::uint32_t RT_DIM            = params.RT_DIM;
#endif
```

when `SPEED_OF_LIGHT` the flag isn't set, argument values are taken from the struct.

In addition to all this, runtime parameters are first copied from `L1` to their core's respective `L0`. The exact location is `stack region`.  Argument to `run_kernel` is encapsulated in `RUNTIME_PARAMS` macro, placed in variant's `build.h`.

### Type of change

- [ ] Documentation update
- [x] Refactoring

### Checklist

### DIFF
--- a/tt_llk_blackhole/llk_lib/llk_math_eltwise_unary_datacopy.h
+++ b/tt_llk_blackhole/llk_lib/llk_math_eltwise_unary_datacopy.h
@@ -26,8 +26,6 @@ inline void _llk_math_eltwise_unary_datacopy_(
     LLK_ASSERT(num_faces == 1 || num_faces == 2 || num_faces == 4, "num_faces must be 1, 2, or 4");
     // For 32bit data, each half of DEST can take 16 tiles. Since dest offset is returned as if 16bit data are used, we need to
     // adjust it to offset in faces for 32bit data.
-    // std::uint32_t dst_index_in_faces = dst_index << 2; // Each tile has 4 faces;
-
     if (unpack_to_dest && is_32bit_input(src_format, dst_format))
     {
         math_unpack_to_dest_math_ready();


tests/python_tests/helpers/test_variant_parameters.py
```diff
@@ -765,21 +758,21 @@ class IN_TILE_DIMS(RuntimeParameter):
 
     def convert_to_cpp(self) -> str:
         lines: list[str] = [
-            f"constexpr int in0_tile_r_dim = {self.in0_r_dim};",
-            f"constexpr int in0_tile_c_dim = {self.in0_c_dim};",
-            f"constexpr int in1_tile_r_dim = {self.in1_r_dim};",
-            f"constexpr int in1_tile_c_dim = {self.in1_c_dim};",
+            f"const std::uint32_t in0_tile_r_dim = {self.in0_r_dim};",
```
fvranicTT:
```suggestion
            f"constexpr std::uint32_t in0_tile_r_dim = {self.in0_r_dim};",
```

Issue #1501, opened 2026-03-18
Bfp4_b in reduce


ldjurovicTT:
Add bfp4_b to reduce functional test ( potentially needed in SDPA )

PR #1503, opened 2026-03-18
Iniital support for bfp4_b as format in reduce tests

ldjurovicTT:
### Ticket
#1501 

### Problem description
Reduce currently doesn't take bfp4_b as input or output format

### What's changed
Added custom test function for bfp4_b as input and output format. Added quantization in golden generation.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### DIFF
--- a/tests/python_tests/helpers/golden_generators.py
+++ b/tests/python_tests/helpers/golden_generators.py
@@ -2094,18 +2094,6 @@ class ReduceGolden:
         elif input_format == DataFormat.Bfp8_b:
             operand = _bfp8b_to_float16b(operand)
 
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_reduce.py
```diff
@@ -75,6 +76,187 @@ def test_reduce(
...
+            res_tensor,
+            formats.output_format,
+            tile_shape=tile_shape,
+            print_errors=True,
+        ), "Assert against golden failed"
+
+
+@parametrize(
+    tile_dimensions=[[32, 32]],
+    formats=input_output_formats(
```
fvranicTT: Can we just narrow this down to Bfp4? And then we don't have to skip the rest.

Issue #1526, opened 2026-03-23
[Test Infra]: Quasar LLK tests broken


atatuzunerTT:
### Description
Commit `88712f37` (PR #1504) breaks Quasar tests.

It refactored BRISC (polling loop, `commit_brisc_command`, `LAST_LOADED_ELFS` / conditional TRISC ELF reload) and removed `reset_mailboxes()` from the start of `run_elf_files()`. It also changed TRISC mailbox layout in `tests/helpers/src/trisc.cpp` and expanded `tests/helpers/include/boot.h` (e.g. unconditional `ckernel_helper.h`, soft-reset / branch-prediction helpers that assume Blackhole/Wormhole cfg and ckernel APIs).

**Quasar has no BRISC**, so several of those assumptions are invalid. We have seen:

- **Build breaks** — Quasar `tt_llk_quasar/common/inc/ckernel.h` vs `tests/helpers/include/ckernel_helper.h` duplicate/conflicting globals; missing `DISABLE_RISC_BP_*` / `get_cfg_pointer` in Quasar’s generated `cfg_defines`.
- **Runtime timeouts** — `wait_for_tensix_operations_finished` calling `commit_brisc_command(RESET_TRISCS)` polls `BriscCounter`, which never updates without BRISC.
- **PCC / golden mismatches** — With `reset_mailboxes()` removed, unpack/math/pack completion mailboxes can still read `KERNEL_COMPLETE` (0xFF) from a **previous** run, so the waiter can return **before** the new kernel finishes and the host reads **stale** L1.

### Approach
I have some preliminary changes in my branch `atuzuner/fix-quasar-trisc-boot-ckernel` for reference, but these, by no means, resolve the issue.

**Typical touchpoints:** `tests/python_tests/helpers/test_config.py`, `tests/python_tests/helpers/device.py`, `tests/helpers/include/boot.h`, `tests/helpers/src/trisc.cpp`, `tt_llk_quasar/common/inc/ckernel.h`

### Impact
Gates https://github.com/tenstorrent/tt-llk/pull/1388 and other Quasar development.

@fvranicTT @ajankovicTT @gdsinghTT fyi

PR #1529, opened 2026-03-24
[infra.] Reintroducing WTF fix with Quasar support. Intruduced multi-core execution for perf

ajankovicTT:
### Ticket

@atatuzunerTT wrote in great detail what was wrong with the inital WTF fix. It is explained here #1526.

### Problem description
Quasar can now execute alongside updated infra. Furthermore, perf tests aren't flaky, thus it is fine to execute them in multi-core fashion. I introduced a flag on CI workflow that enables us to selectivelly turn multi-core execution on. Perf runtime is now performed on one runner for each card, taking ~15 minutes on WH and around ~25 minutes on BH.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tests/python_tests/helpers/llk_params.py
+++ b/tests/python_tests/helpers/llk_params.py
@@ -352,13 +352,6 @@ class MailboxesPerf(Enum):
     BriscCounter = Unpacker + 20
 
 
-class MailboxesPerfQuasar(Enum):
-    Unpacker = 0x1FFC4
-    Math = 0x1FFC8
-    Packer = 0x1FFCC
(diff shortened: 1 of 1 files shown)


tests/helpers/src/brisc.cpp
```diff
@@ -4,26 +4,142 @@
...
+    }
+}
+
+#else
+
+int main()
+{
+}
+
+#endif
```
nvelickovicTT: Same here, as it is difficult to follow these since they don't have indentation.

Issue #1541, opened 2026-03-25
[Quasar] Batched unpack-to-dest tilize (_llk_unpack_tilize_block_)


skotaracTT:
Why block tilize is needed: 32-bit formats (Int32, Float32) use the unpack-to-dest path — data goes directly from L1 to the dst reg bypassing the math thread entirely. The previous implementation tilized one tile at a time, calling section_done after each tile. This forced the packer to drain dst after every single tile, creating a tight ping-pong between unpack and pack threads.

Block tilize places all tiles into dst in one shot (up to 8 for 16-bit dst, 4 for 32-bit), then signals section_done once. The packer reads all tiles in a single batch. This eliminates per-tile synchronization overhead. The L1 and dst counters are programmed once per tile row, and the MOP hardware loop handles the rest.

PR #1543, opened 2026-03-25
[Quasar] Implement _llk_unpack_tilize_block_ (perf gain)

skotaracTT:
### Ticket

https://github.com/tenstorrent/tt-llk/issues/1541
https://github.com/tenstorrent/tt-llk/pull/1521#discussion_r2984710589
### What's changed

- Implement `_llk_unpack_tilize_block_`API for batched unpack-to-dest tilize on Quasar
   - `_llk_unpack_tilize_block_init_`: Same stride register setup as the non-block init, except DST_Z_STRIDE = NUM_FACES * 16 so each Z increment skips one full tile in dst reg.
   - `_llk_unpack_tilize_block_mop_config_`: Single-instruction MOP with Dst_Z_Cntr_inc=1 and Src_Z_Cntr_inc=1. No dvalid per tile (packer waits for section_done). Last tile uses a reset variant (Cntr_Reset_Mask=1) to clean up counter state.
   - `_llk_unpack_tilize_block_`: Sets L1 source and dst face counters once per tile row, then runs the MOP which processes BLOCK_CT_DIM tiles, auto-advancing both L1 and dst positions.
- Place all tiles into dst in one shot, signal `section_done` once, pack in a single batch
- Fix stale dst face counter leaking between unpack-to-dest and normal tilize paths                  
   - `_llk_unpack_tilize_` (existing, fix): Added SET_DST(FACE_SEL, 0) to clear the dst face counter. The block path leaves this counter non-zero, which corrupted subsequent non-block runs (e.g. Int16 after Int32) in the same simulator session.

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### DIFF
--- a/tests/sources/quasar/unpack_tilize_quasar_test.cpp
+++ b/tests/sources/quasar/unpack_tilize_quasar_test.cpp
@@ -29,25 +29,22 @@ void run_kernel(RUNTIME_PARAMETERS params)
     constexpr auto dest_producer = unpack_to_dest ? dest_dvalid_client::UNPACK : dest_dvalid_client::FPU;
     set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_producer, dest_dvalid_client::PACK});
 
-    if constexpr (unpack_to_dest)
+    if constexpr (unpack_to_dest && is_fp32_dest_acc_en)
     {
-        if constexpr (is_fp32_dest_acc_en)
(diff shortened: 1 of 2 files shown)


tests/python_tests/quasar/set_simulator_env.sh
fvranicTT: Should we remove this file?

PR #1562, opened 2026-03-30
[Quasar AI codegen] Add generated abs SFPU kernel for Quasar (Opus)

vvukomanovicTT:
Generated via LLK codegen (Opus). Includes kernel implementation, SfpuType enum entry, and test files.

### Problem description
Quasar is missing an SFPU `abs` (absolute value) kernel. The Blackhole architecture already has this implementation, but it has not been ported to Quasar.

### What's changed
- Added `tt_llk_quasar/common/inc/sfpu/ckernel_sfpu_abs.h` — Quasar SFPU absolute value kernel using SFPABS instruction
- Added `tt_llk_quasar/llk_lib/llk_defs.h` — `abs` entry in SfpuType enum
- Added `tests/sources/quasar/sfpu_abs_quasar_test.cpp` — C++ test source for abs kernel
- Added `tests/python_tests/quasar/test_sfpu_abs_quasar.py` — Python test harness for abs on Quasar simulator

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tests/sources/quasar/sfpu_abs_quasar_test.cpp
+++ b/tests/sources/quasar/sfpu_abs_quasar_test.cpp
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
-// AI-generated — run_id: 2026-04-01_abs_quasar_779f878d
+// AI-generated — run_id: 2026-04-08_abs_quasar_2f52d870
 
 #include <cstdint>
 
@@ -27,12 +27,18 @@ void run_kernel(RUNTIME_PARAMETERS params)
 
     if (unpack_to_dest)
     {
-        // Unpacking to DEST directly
+        // Direct path: UNPACK writes data straight into Dest — no FPU datacopy needed.
+        // Requires format bit-width to match Dest mode (e.g., 16-bit input with 16-bit Dest).
+        // dvalid clients: UNPACK (writes Dest), SFPU (reads/writes Dest), PACK (reads Dest).
         set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_dvalid_client::UNPACK, dest_dvalid_client::SFPU, dest_dvalid_client::PACK});
+        // When unpacking directly to Dest (bypassing FPU), MATH HW still needs format configuration
+        // so the SFPU reads Dest data in the correct format.
         _llk_math_upk_to_dest_hw_configure_<IMPLIED_MATH_FORMAT, is_fp32_dest_acc_en, false /*is_int_fpu_en*/>();
     }
     else
     {
+        // FPU path: UNPACK -> SrcA -> FPU datacopy (MOVA2D) -> Dest.
+        // Needed when input format bit-width differs from Dest mode (format conversion required).
         set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_dvalid_client::FPU, dest_dvalid_client::SFPU, dest_dvalid_client::PACK});
     }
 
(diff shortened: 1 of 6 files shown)


tests/sources/quasar/sfpu_abs_quasar_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+    else
+    {
+        set_up_dest_dvalid_per_thread<dest_dvalid_client::UNPACK>({dest_dvalid_client::FPU, dest_dvalid_client::SFPU, dest_dvalid_client::PACK});
+    }
+
+    buffer_descriptor_u bd_val = {0};
+
+    bd_val.f.l1_addr_16B = L1_ADDRESS(params.buffer_A[0]);
+    bd_val.f.format      = static_cast<std::uint8_t>(formats.unpack_A_src);
+    bd_val.f.x_dim       = params.TEST_FACE_C_DIM;
```
nvelickovicTT: Maybe an overkill, but I think we should force it to comment on such assignments, as these are easy to just copy paste without thinking much and making a mistake. It's easy for humans to.
A comment summarizing values of x, y and z dim would be a good way for the reviewer to more easily catch it.
For example I would expect: `In the buffer descriptor X dim represent columns, Y dim rows in a face, where Z dim is number of faces.`

tt_llk_quasar/llk_lib/experimental/ckernel_sfpu_abs.h
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
nvelickovicTT, line 1: Am I missing something or it created 2 exactly same files in 2 different locations? Do we need this? Seems wrong.

tests/sources/quasar/sfpu_abs_quasar_test.cpp
```diff
@@ -0,0 +1,201 @@
...
+    {
+        // FPU path: data goes UNPACK -> SrcA -> FPU (MOVA2D) -> Dest -> SFPU -> PACK
+        set_up_dest_dvalid_per_thread<dest_dvalid_client::FPU>({dest_dvalid_client::FPU, dest_dvalid_client::SFPU, dest_dvalid_client::PACK});
+        set_up_dest_dvalid_per_thread<dest_dvalid_client::SFPU>({dest_dvalid_client::FPU, dest_dvalid_client::SFPU, dest_dvalid_client::PACK});
+    }
+
+    DataFormat src_format = static_cast<DataFormat>(formats.math);
+    _llk_math_srcAB_hw_configure_<IMPLIED_MATH_FORMAT, is_fp32_dest_acc_en, is_int_fpu_en>(src_format, src_format);
+
+    const std::uint32_t num_sfpu_iterations = params.TEST_FACE_R_DIM / ckernel::math::SFP_ROWS;
```
nvelickovicTT, line 121: So, this should be number of needed iterations per face, right? But, in the math call bellow it iterates on TILE_CNT. It looks like it will do TILE_CNT faces, and not TILE_CNT tiles, no?
I might be missing something here, but I am confused.

tests/sources/quasar/sfpu_abs_quasar_test.cpp
```diff
@@ -0,0 +1,201 @@
...
+            _llk_math_eltwise_unary_datacopy_(num_rows, i);
+        }
+
+        _llk_math_set_dvalid_<p_cleardvalid::FPU>();
+    }
+
+    _llk_math_eltwise_unary_sfpu_init_();
+
+    // Apply SFPU absolute value (SFPABS instruction) to all tiles
+    for (std::uint32_t i = 0; i < params.TILE_CNT; ++i)
```
nvelickovicTT, line 141: This seems wrong, let's review if it is indeed, and if yes whether its AI problem or existing tests problem.

Because num_faces is not used anywhere.

tests/sources/quasar/sfpu_abs_quasar_test.cpp
nvelickovicTT: So, I did miss the fact that on quasar it only operates in VectorMode::RC. It's the default and only option so it is not set explicitly. Which is why by iterating over 1 face, it will automatically iterate over the whole tile.

tests/sources/quasar/sfpu_abs_quasar_test.cpp
nvelickovicTT: It is actually fine, although not immediately clear to the reader what happens there.
I wonder if you prompt it to include in the comment the iterating strategy, what it would say there. As the original code also does not explain it.

PR #1582, opened 2026-04-01
[infra.] Recording, reproducing and processing test variant execution order

ajankovicTT:
### Problem description
We have a problem where some tests fail if they go before some specific tests. This PR introduced order tracking mechanism where the exact order in which variants ran is recorded to a json file, which can later be used to replay them in that exact same order, to faitfully be able to reproduce the bugs that emerge as a result of hardware state leakage. Furthermore, the new python sciprt is introduced is used to process these test order files.

### What's changed
Added flags:

| Argument | Description |
|:----|:----|
| `--record-test-order[=./path/to/dest/file.json]` | Tracks which test variants executed on what core alongside dumping `Tensix` state after each variant finished. Default path is `./tt-llk/../run_order_[month]_[day]_[hour]_[minute]_[second].json` |
| `--test-order-file=./path/to/order/file.json` | Tests names present in the file are executed in the exact order in which they are written. This feature is still work in progress |
| `--rewind-runner='runner_name'` | what is the runner key that should be used when processing variant runs in a file |

and acompanying changes to enable their funcioning.

reduced number of `pytest` runners in setup-and-test.yml from 15 to 10 to improve execution stability.

### Type of change

- [x] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tests/python_tests/helpers/test_config.py
+++ b/tests/python_tests/helpers/test_config.py
@@ -1237,6 +1237,8 @@ class TestConfig:
                 device_module.Mailboxes.BriscCommand0,
                 device_module.Mailboxes.BriscCommand1,
                 device_module.Mailboxes.BriscCounter,
+                device_module.Mailboxes.BriscBread0,
+                device_module.Mailboxes.BriscBread1,
             }
         test_target = TestTargetConfig()
         timeout = 600 if test_target.run_simulator else timeout


tests/python_tests/helpers/utils.py
```diff
@@ -265,6 +265,9 @@ def _bfp4_block_aware_compare(
     return is_valid
 
 
+_RECORD_TEST_ORDER: bool = False
```
fvranicTT: This is always false, as it is not tied to the same variable from `conftest.py`.

tests/python_tests/conftest.py
```diff
@@ -16,12 +21,24 @@
...
+from ttexalens import check_context, tt_exalens_init
+from ttexalens.tt_exalens_lib import get_tensix_state
 
 _exalens_server: Optional[ExalensServer] = None
 
 
+# This is a workaround for this issue: https://github.com/tenstorrent/tt-exalens/issues/958
+# In a nutshell, everything except Tensix GPRs is accessible over NoC, thus ignoring that allows us to dump
+# most of the Tensix state, without causing any runtime issues.
+def overrirde_gprs_used_by_tensix_dump():
```
fvranicTT:
```suggestion
def override_gprs_used_by_tensix_dump():
```

tests/python_tests/helpers/golden_generators.py
```diff
@@ -144,6 +149,108 @@ def get_golden_generator(cls):
...
+                        GeneratorProxy.STIMULI_CACHE_ROOT / stimuli_id / "golden.pt"
+                    )
+                    result = torch.load(golden_path)
+
+                elif GeneratorProxy.MODE == ProxyMode.CACHE_GOLDEN:
+                    result = attr(*args, **kwargs)
+                    # We cache tensor value in TEMP_RESULT when we call Stimuli_Config.save_to_caches
+                    GeneratorProxy.TEMP_RESULT = result
+
+                return result
```
fvranicTT: If GeneratorProxy mode is neither result is unbound.

Issue #1141, opened 2026-01-21
Optimize SFPU based reductions.


amahmudTT:
### Optimize SFPU based reduce

The SFPU based reductions can be optimized by a somewhat non-trivial percentage.  Here are the steps. 

For column reduce, we could remove one transpose only. we would still be summing up the elements in a column but in a different order like instead of adding elements from row 0 with 1 , 2, 3, you probably would be adding 0, 4, 8, 12.

For Wormhole, some of the nops required after some SFPU instructions can be removed by reordering the order of operations. 

For columnwise reduction 

We transpose 
we are doing a reduction across the registers (0 to 3 and result in 3) and (4 to 7 and result in 4) 
We transpose
again we are doing a reduction across the registers (0 to 3 and result in 3) and (4 to 7 and result in 4)
We add register 0 with 4

But we could just do 

 a reduction across the registers (0 to 3 and result in 3) and (4 to 7 and result in 4) 
 add register 0 with 4 
 transpose 
 a reduction across the registers (0 to 3 and result in 3)

In the previous approach the addition of 0, and 4, the first lanes computation matters only, in this approach when we add 0, and 4 useful addition is being done across all the lanes, thus reducing the total operations.

PR #1628, opened 2026-04-09
Optimize SFPU column reductions with Reduce→Add→Transpose→HalfReduce …

nstamatovicTT:
…algorithm (#1141)

Replace the naive Transpose→Reduce→Transpose→Reduce→Add column reduction with an optimized Reduce→Add→Transpose→HalfReduce approach that eliminates one transpose and halves the second reduction pass.

Key changes:
- Remove sum_columns() helper in favor of inline optimized sequence
- Record two replay buffers: full reduce (positions 0-5) and half reduce (positions 6-8 on BH, 6-10 on WH with NOPs)
- Interleave A-group (LREG0-3) and B-group (LREG4-7) operations to hide pipeline latency, eliminating 6 NOPs from WH float reduce
- Simplify row reduce to always use replay(0, 6) since the interleaved buffer works for both int and float

Instruction savings per column reduce iteration:
- BH int/float: 16 instructions saved
- WH int: 16 instructions saved
- WH float: 52 instructions saved (NOP elimination + algorithm improvement)

Tested: 880/880 sfpu_reduce + 1/1 sfpu_reduce_sdpa + 4284/4284 reduce variants pass on Blackhole hardware. WH compiles cleanly (no HW available).

### Ticket

[Issue](https://github.com/tenstorrent/tt-llk/issues/1141)

### Type of change

- [ ] Documentation update
- [ ] Refactoring

### Checklist

### DIFF
--- a/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
+++ b/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_reduce.h
@@ -95,21 +95,6 @@ inline void load_face_data(std::uint32_t upper_face_addr, std::uint32_t lower_fa
 }
 
 /**
- * @brief Perform column-wise summation using transpose and replay buffer
- *
- * Process column sums for both upper and lower face using transpose and replay buffer.
- * @tparam replay_buffer_length The replay buffer index to use (6 for both int and float on Blackhole)
- */
-template <std::uint32_t replay_buffer_length>
-inline void sum_columns()
-{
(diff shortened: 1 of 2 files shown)


nvelickovicTT (approved):

nvelickovicTT:
This looks really good!
I have couple of questions:
- Do you have results of perf tests? Is the instruction saving just counting instructions directly, or also the implicit NOPs for BH, for example?
- Would it be too cluttered if we made AI paste instructions from replay buffer as comments wherever it does `lltt:replay`? It is difficult to read that part. Another option would be to record it in the first execution instead of separately. It would make it more readable, and maybe overall more performant slightly.

amahmudTT (approved):

